PR #8, opened 2017-02-03
CMakeTweak

apmccartney:
+ Fortran flags play nice with mixed-language projects
+ Silence warnings when subproject
+ More aggressive link time optimization when compiling executable
+ Removed erroneous Intel compiler flags for release and native configurations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ os:
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install gfortran; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install gcc; fi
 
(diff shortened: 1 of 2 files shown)


apmccartney:
Sure, but first, understand that modern CMake allows for compiler flags to be set on a project-by-project basis, as opposed to the way we used to, e.g.
```
# this setting applies to all C++ projects
set( CMAKE_CXX_FLAGS "-fmy-cool-flag" ) 

# this adds another flag specifically to myTarget
target_compile_options( myTarget PRIVATE -fmy-other-flag )
```

Let's assume that `myTarget` is a library. The difference between the `PRIVATE` and `PUBLIC` parameter in this case is that, for PUBLIC compile options, the option is also added to any target linking to `myTarget` (i.e. it's transitive), where for PRIVATE compile options,  the option is not transitive. 

I haven't determined the implications to link-time optimization yet. If it turns out that it's an orthogonal concern, I'll modify Metaconfigure to use `Private` compile options by default.

PR #11, opened 2017-03-02
Changing physical constant values.

jlconlin:
Surprisingly, this changed the outputs in very significant ways. How do we know which one is correct?

### DIFF
--- a/src/phys.f90
+++ b/src/phys.f90
@@ -18,7 +18,7 @@ implicit none
   real(kr),parameter,public::amassn=1.00866491588_kr
   real(kr),parameter,public::amu=1.660539040e-24_kr
   real(kr),parameter,public::hbar=1.054571800e-27_kr
-  real(kr),parameter,public::ev=1.6021766208e-19_kr
+  real(kr),parameter,public::ev=1.6021766208e-12_kr
   real(kr),parameter,public::clight=2.99792458e10_kr
 
 ! CODATA 1986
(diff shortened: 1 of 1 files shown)


apmccartney:
Sorry for the delay. In which test+tapes did you find non-negligible changes?

src/phys.f90
```diff
@@ -14,20 +14,20 @@ module physics
...
-  real(kr),parameter,public::bk=8.617385e-5_kr
-  real(kr),parameter,public::amassn=1.008664904e0_kr
-  real(kr),parameter,public::amu=1.6605402e-24_kr
-  real(kr),parameter,public::hbar=1.05457266e-27_kr
-  real(kr),parameter,public::ev=1.60217733e-12_kr
+! real(kr),parameter,public::bk=8.617385e-5_kr
+! real(kr),parameter,public::amassn=1.008664904e0_kr
+! real(kr),parameter,public::amu=1.6605402e-24_kr
+! real(kr),parameter,public::hbar=1.05457266e-27_kr
+! real(kr),parameter,public::ev=1.60217733e-12_kr
```
apmccartney: why did ev change by 7 orders of magnitude?

kahlerac: Also, based upon what I started, but never completed for NJOY2012 a while ago there are also a number of particle mass constants to update ... broadr, heatr, groupr, thermr, leapr at least.

apmccartney:
Following the electronVolt value correction, which test/tape combinations had significant changes?

jlconlin:
With all of the constants changed, only tests 3,5, and 6 passed. I have not checked to see specifically why the others failed.

PR #17, opened 2017-04-13
Upn12

jlconlin:


apmccartney:
I'd like the have some confidence the issue was resolved. The flags on master won't catch some of the issues we're seeing on new_meta

Let's update the GCC debug flags and rerun

apmccartney:
well done

PR #19, opened 2017-05-04
Changes to run ENDF/B-VIII.beta4 LEAPR inputs

marquezj:
This pull request fixes two issues that appeared while testing the LEAPR input files available in the GForge sever:

https://ndclx4.bnl.gov/gf/project/endf/scmsvn/?action=browse&path=/trunk/endf7/thermal_scatt/

* For tsl-SiO2-alpha.leapr and tsl-SiO2-beta.leapr (or any input with secondary scatterers), the array tempr has to be deallocated before processing the secondary scatterer.

* For tsl-ortho-H.leapr, tsl-para-H.leapr, tsl-ortho-D.leapr and tsl-para-D.leapr there is a conflict between the coldh and skold subroutines. Both read and use S(kappa), but when the skold subroutine was implemented in NJOY99 it did not coexist nicely with the (previous) coldh subroutine. With these changes both the H2/D2 LEAPR inputs (that use coldh) and the O(D2O)/D(D2O) inputs (that use skold) run without errors.


jlconlin:
@marquez Do you have some specific NJOY inputs that demonstrates the problem as well as the problem being fixed?

marquezj:
Sure, 

For the first bug, the attached input file tsl-SiO2-alpha.leapr gives:

Fortran runtime error: Attempting to allocate already allocated variable 'tempr'

For the second bug, the attached input file tsl-ortho-H.leapr gives a Segmentation fault.

Both were downloaded from the ENDF/B-VIII [GForge](https://ndclx4.bnl.gov/gf/project/endf/scmsvn/?action=browse&path=/trunk/endf7/thermal_scatt/) server. The first bug also appear with  tsl-SiO2-beta.leapr, and the second bug with  tsl-ortho-D.leapr,  tsl-para-H.leapr and  tsl-para-D.leapr.



[tsl-ortho-H.leapr.zip](https://github.com/njoy/NJOY2016/files/977005/tsl-ortho-H.leapr.zip)
[tsl-SiO2-alpha.leapr.zip](https://github.com/njoy/NJOY2016/files/977012/tsl-SiO2-alpha.leapr.zip)

jlconlin:
Thanks! I'll take a look and see what we find.

src/leapr.f90
```diff
@@ -293,7 +295,7 @@ subroutine leapr
 
    !--allocate storage for ssm (and ssp if needed)
    allocate(ssm(nbeta,nalpha,ntempr))
-   if (nsk.ne.0) allocate(ssp(nbeta,nalpha,ntempr))
+   if (ncold.ne.0) allocate(ssp(nbeta,nalpha,ntempr))
```
jlconlin: @marquezj What is the significance of changing from testing `nsk` to `ncold`?

marquezj: nsk controls the use of the Skold (nsk = 2) or Vineyard approximation (nsk = 1, not implemented). ncold controls the use of the Young-Koppel approximation to H2/D2 scattering. The S(a,b) for H2 and D2 computed in coldh is not symmetric and that is why you need to allocate the array ssp for positive beta.

TL;DR: ssp is used in coldh() and not in skold().

jlconlin:
I have run those LEAPR input decks in both versions (ours and @marquezj). This pull request does make those input decks not fail, but it is difficult to know if the changes are correct.

marquezj:
* You can compare that, with the proposed change, tsl-SiO2-alpha.leapr and tsl-SiO2-beta.leapr reproduce the MF=7,MF=4 section of the ENDF/B-VIII.0-beta4. The input files do not reproduce the MF=7, MF=2 section of the evaluation because a separate issue that (I think) will be discussed at mini-CSEWG.

* Comparing the evaluations for ortho/para hydrogen/deuterium is a bit harder because the reference evaluations are very old, and there might be different round off errors from different compilers/platforms. But, you can compare the result of running these LEAPR inputs with NJOY99.0 (NJOY99 w/o updates) and with NJOY2016.0 compiled in the same computer/compiler.

marquezj:
This is a detailed description of the above tests:

- To test the first change, run the `tsl-SiO2-alpha.leapr` input file:
`$ ../build/njoy < tsl-SiO2-alpha.leapr `
Remove MF=7,MT=2 and line numbers:
`$ cat tape24 | grep -v " 7  2" | cut -b 1-75 > temp1`
Do the same thing with reference ENDF-6 file:
`$ cat tsl-SiO2-alpha.endf.original | grep -v " 7  2" | cut -b 1-75 > temp2`
and compare with diff:
`diff temp1 temp2`
The only differences in MF=7,MT=4 are two numbers (B(8), B(12), in ENDF-103 Sect. 7.4) changed by hand in the evaluation, and explained in the README file (tsl-SiO2-alpha.readme):

>   4) To reproduce the published ENDF/B-VIII library
     the cross section and number of secondary scatterers
     must be changed to 3.74875 and 2 in the File 7 
     generated from the provided input file.

Same thing happens with  `tsl-SiO2-beta.leapr`.

- For the second change, run the `tsl-ortho-H.leapr` input file:
`$ ../build/njoy < tsl-ortho-H.leapr`
Remove line numbers:
`$ cat tape20 | cut -b 1-75 > tsl-ortho-H.endf.njoy16`
Redo with NJOY99 up0 (we compare with this version because it is before the IKE group introduced the Skold subroutine):
`$ ../njoy99.up0/njoy < tsl-ortho-H.leapr`
`$ cat tape20 | cut -b 1-75 > tsl-ortho-H.endf.njoy99`
Compare with diff:
`$ diff tsl-ortho-H.endf.njoy16 tsl-ortho-H.endf.njoy99`
Files are the same. Same thing happens with `tsl-para-H.leapr`, `tsl-ortho-D.leapr`, and `tsl-para-D.leapr`.

jlconlin:
@marquezj That is great information. Thank you for taking the time to write up that description.

We would like to add the two inputs you provided as tests in our suite of tests. Can you please create them and add them to your branch before we do this pull request? We are in the process of adding Test21 in a different branch. Can you please create Test22 and Test23? Take a look at how the other tests are done and duplicate the process. Or, you can see Test21 we are creating in [Pull Request 21](https://github.com/njoy/NJOY2016/pull/21).

jlconlin:
@marquez I see you have been busy adding the tests for this Pull Request. Did you intend to put them on a separate request?

Also note that we have updated our master branch. It shouldn't cause any conflicts when you merge into your branch.

marquezj:
Yes, I created the test in a different branch, and created a pull request for just the test  (pull request #23 ). 

But first I tried it my master branch and made the mess you see here.

jlconlin:
@marquezj I don't think you have the latest updates in that branch. It also seems like you don't have the same changes that you have created in this Pull Request. That would certainly make the tests fail.

It may just be easier to add the tests to this Pull Request instead of creating a separate one.

marquezj:
Ok. I thought you wanted to merge the test first and then the patch.

Now it is here.

jlconlin:
Sorry. I should have said it should go together.

jlconlin:
@marquezj What is the status on this Pull Request? I'd like to close this out if you could add the second test we discussed at WPEC.

docs/testDescription.md
```diff
@@ -121,3 +121,10 @@ title: NJOY2016 Test Descriptions
 [[input](https://raw.githubusercontent.com/njoy/NJOY2016/master/test/20/input)]
 
   This test case uses a preliminary ENDF/B-VII.1 evaluation for <sup>35</sup>Cl, which uses the advanced Reich-Moore-Limited format for the resonances.  This allows more complex covariances to be represented.  In this case, covariances involving the (n,p) reaction are calculated.
+
+## Test Problem 21
```
jlconlin: @marquezj Please add descriptions of the two new tests that you have added.

marquezj:
Done.

src/leapr.f90
```diff
@@ -130,6 +133,7 @@ subroutine leapr
    !    za      1000*z+a for principal scatterer
    !    isabt   sab type (0=s, 1=ss, def=0)
    !    ilog    log flag (0=s, 1=log10(s), def=0)
+   !    smin    minimum S(alpha, beta) stored in file (def=2e-38)
```
jlconlin: I like that the value of `smin` is a user option. This is a good improvement. However, we are going to change the default to 1.0E-75 to preserve the double precision of the value. The change in the default value is going to be back-ported to NJOY2012, but we will not provide a user option to change it.

We can make the change after this pull request is finished, but it would be easier if it was part of this pull request.

@marquezj What would you like to do here?

marquezj: Regarding smin, for the time being we recommend the use of 2E-38, and rolled back the change introduced by Mattes and Keinert in NJOY99-up114. This minimum value of S(alpha,beta) was reduced in up114 without changing the other tolerances, and produced artifacts in the evaluation of some of our thermal scattering libraries for cold moderators. We understand that 64 bits machines would call for better precision in the calculation of thermal scattering libraries, but this is something that should be done taking into account all the quantities that affect precision in LEAPR (small, vsmall, tiny, smin, eps, nphon, etc) and not just the cut off when the ENDF-6 library is prepared. 

That said, if you want to preserve the double precision of the value: why not use something closer to the minimum value of double precision (2.2250738585072013E-308) instead of an arbitrary intermediate value?

PR #33, opened 2017-09-21
Modified moder to correctly read the new mf6 mt18 structure

whaeck:
Following the change in the structure of mf6 mt18, moder was modified as follows:
- code for old negative laws is removed because they are not compatible with the new structure
- check for negative law only allowed for fission

The correction was tested on 2 files provided by Ionel Stetcu (one case for JP=11 and another one for JP=21) and they now pass the moder run without an error message stating that the law is illegal.

### DIFF
--- a/src/moder.f90
+++ b/src/moder.f90
@@ -887,6 +887,7 @@ contains
    real(kr)::a(*)
    ! internals
    integer::nk,ik,law,ne,ie,nmu,imu,nb,nw
+   character(60)::strng
 
    nk=n1h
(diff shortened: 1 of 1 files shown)


jlconlin:
This is a pretty simple change. I'd like to add a test so that we know this is doing as expected. However, at the same time, this isn't an official format change (yet) so I'm also reluctant to add another test that may not be official.

whaeck:
You can put that in the issue tracker so that we do not forget about adding the test.

PR #34, opened 2017-09-22
Updated moder to take into account format changes in MF1 MT458

whaeck:
MF1 MT458 can now contain energy dependent tabulated data for the components of fission energy release.

Moder was modified to take into account this change so that NJOY can read the ENDF file without giving an error message stating that a SEND record was expected.

The NJOY manual will be modified to reflect the addition of an error message.

### DIFF
--- a/src/moder.f90
+++ b/src/moder.f90
@@ -1437,7 +1437,7 @@ contains
                     call mess('file32','1illegal value of ndigit',&
                               'set default, ndigit=2')
                   else
-                    call error('file32','1illegal value of ndigit',' ')
+                    call error('file32','illegal value of ndigit',' ')
                   endif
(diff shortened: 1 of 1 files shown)


jlconlin:
Where is the documentation for this new format? The code looks reasonable, but I don't know how it matches the format.

whaeck:
You should check the NNDC website and look for the ENDF manual - development version: http://www.nndc.bnl.gov/endf/format/endf-manual-latest.pdf

When I checked MT458 yesterday evening, I checked the wrong file (it did not contain the additional TAB1 records).

jlconlin:
This is fairly straightforward. It just reads over the data, but does nothing with it (for now).

PR #42, opened 2017-10-20
Feature/groupr endf8

jlconlin:
This patch allows us to process U-235 and U-238 through GROUPR without crashing.

### DIFF
--- a/src/groupr.f90
+++ b/src/groupr.f90
@@ -4698,6 +4698,7 @@ contains
    real(kr),parameter::emax=1.e10_kr
    real(kr),parameter::zero=0
    save lnu,ip,ir
+   character(60)::strng
 
    !--initialize
    if (e.gt.zero) go to 200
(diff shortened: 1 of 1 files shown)


src/groupr.f90
```diff
@@ -4753,7 +4754,8 @@ subroutine getyld(e,enext,idis,yld,mat,mf,mt,lfs,itape)
    call skip6(itape,0,0,tmp(loc),law)
   170 continue
    if (ik.lt.nk) go to 130
-   call error('getyld','requested nuclide not found',' ')
+   write(strng,'(''unable to find nuclide for iza='',i7,'' lfs='',i3)') iza,lfs
```
whaeck: Maybe change this to
write(strng,'(''unable to find nuclide for iza='',i7,'' lfs/isom='',i3)') iza,lfs
because in some cases, we are looking for the isomeric state and in others for the excited level.

Issue #10, opened 2017-02-24
CTest on Windows fails


jlconlin:
This is because the python can't remove the `output` and `error` file.

PR #43, opened 2017-10-25
Fix Windows ctest failure.

jlconlin:
This fixes issue #10.


jlconlin:
@whaeck Can you take a look at this?

PR #45, opened 2017-12-05
Windowsy linux

nicholas-sly:
Added appveyor.yaml file that builds several versions of NJOY2016 on Windows.  Currently uses MinGW, MSYS, and Cygwin.  Cygwin does not pass the unit tests, but it was decided by [Austin McCartney](@apmccartney)  that the differences appeared to be acceptably small.  The appveyor control panel will have to be set up to track the master branch once this file is merged in.

Also includes the one additional file from the docker branch.  This Dockerfile can be used by Docker to build an image that includes a working version of the master branch of NJOY2016.  While a viable option to have users run this, it is advised that we should generate a curated Docker image and put it on [Docker Hub](https://hub.docker.com/) (free), as this would make obtaining a working version of NJOY2016 even simpler.  As an example, the command
```bash
docker run -ti NJOY2016
```
will download the image, if it hasn't been downloaded before, and start the container returning the user to a command line inside of the container.  Once there, the user will be able to run all of the tests and enter their own NJOY2016 inputs to run.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,5 +1,5 @@
 if ( NOT PYTHON_EXECUTABLE )
-  find_package( PythonInterp 3.4 )
+  find_package( PythonInterp )
   if ( NOT PYTHONINTERP_FOUND )
     message( FATAL_ERROR "Python interpeter installation was not found." )
   endif()


metaconfigure/fetch_subprojects.py
```diff
@@ -30,8 +30,8 @@ def clone_submodule( relative_path ):
       ----------------------------------------
       """.format( relative_path = relative_path ) ) )
     invocation = [ "git", "submodule", "update", "-q","--init", "--", relative_path ]
-    if os.name == "nt":
```
jlconlin: This should be a change that is made in metaconfigure, which should be pulled into here.

nicholas-sly: I wanted to see if there were any projects that actually needed this before removing it from metaconfigure.  My testing on the VMs and Appveyor show that these commands work on the command prompt, at least for this project.

tests/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 if ( NOT PYTHON_EXECUTABLE )
-  find_package( PythonInterp )
+  find_package( PythonInterp 3.4 )
```
jlconlin: Please remove this restriction. I know we state that 3.4+ is required, but it isn't (yet). I don't want someone's build to break right before NEDF/B-VIII.0 is released.

PR #49, opened 2017-12-15
Feature/heatr-endf8

whaeck:
The NJOY HEATR module has been updated to be able to use the MF1 MT458 section which can now contain tabulated fission q components. This is a new feature in ENDF/B-VIIIb6.

There are now three format types for the fission q components:
- thermal point (this is the historical way of doing fission q components)
- polynomial (introduced for ENDF/B-VII)
- tabulated (introduced for ENDF/B-VIII)

The implementation in the HEATR module was changed so that all types are used consistently. Fission heating is now calculated as:

`H = [ QFR(E) + QGP(E) ] * sigma`

for all types. `QFR` is the kinetic energy of the fission products and `QGP` is the prompt gamma contribution. As a result, evaluations using thermal point will give slightly different heating values because the implementation has been changed (differences of the order of 0.1% can be observed for most energies, at energies above 10 MeV, this can be around 1 %). The polynomial type still yields the same values (the implementation has changed somewhat but the actual formula that is used was not).

The NJOY version number was also changed to 21 to distinguish this version from the previous one as the fully ENDF/B-VIII compatible one.

### DIFF
--- a/src/vers.f90
+++ b/src/vers.f90
@@ -4,6 +4,6 @@ module version
    implicit none
    private
    character(8),public::vers='2016.21'
-   character(8),public::vday='15Dec17'
+   character(8),public::vday='08Jan18'
 end module version
 
--- a/src/heatr.f90
+++ b/src/heatr.f90
@@ -446,7 +446,7 @@ contains
    integer::mfd,mtd,j,idone,mf4,mt4,ie,nmu,imu,npkk,idis
    integer::ielem,ik,nr,idnx,nen,law
    integer::lrel,nmod
-   integer::i6t,nl
+   integer::i6t,nl,l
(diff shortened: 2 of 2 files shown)


src/vers.f90
```diff
@@ -3,7 +3,7 @@ module version
    ! These values are updated during the NJOY revision-control process.
    implicit none
    private
-   character(8),public::vers='2016.20'
-   character(8),public::vday='30Jul17'
+   character(8),public::vers='2016.21'
```
jlconlin: 👍

whaeck:
Tests with the pu239 file from the ENDF/B-VIIIB6 beta has shown that the current implementation has issues reading the tabulated components when they are large in size (over 3000 points). This pull request is on hold till I can fix this.

PR #53, opened 2018-01-09
Feature/heatr endf8

whaeck:
Bug fix: when running heatr twice, an error is generated because the afr array is allocated a second time without being deallocated. This has been corrected.

Question: do we change the date and/or version number for this minor fix?


jlconlin:
Perhaps we should merge this into master and only increment the number when the ERRORR fix is implemented. Thoughts?

whaeck:
This will be pulled into another branch prior to being pulled into master.

PR #54, opened 2018-01-10
Feature/heatr endf8

whaeck:
Small fix: deallocating array at the end of heatr to allow heatr to run correctly when it is run a second time (otherwise you get an error for allocating an already allocated array).


jlconlin:
Approved. Merging into branches other than the master do not require a formal pull request.

PR #56, opened 2018-01-11
Feature/tests

whaeck:
Adding two tests:
- test 24: testing the new tabulated fission energy release components and double heatr run capabilities
- test 25: producing continuous energy thermal scattering ACE files (three files are produced, one for each iwt value in the ACER input)

Three new ENDF files were added from ENDF/B-VIII.0b6: Pu239, H1 and H in H2O thermal scattering.

The current tests are fully compatible with NJOY 2016.22. Test 25 will change when changes made by D. Roubtsov for thermal scattering will be released (only iwt=2 changes).


whaeck:
Included in the fuzzy diff pull request.

PR #58, opened 2018-01-18
Feature/heatr endf8

whaeck:
Fixed the issue for heatr when using ENDF files without an MT458 section (mea culpa).
Such files now give the same result as before (and as NJOY 2012).

A test was added to detect this in the future.


src/heatr.f90
```diff
@@ -757,6 +755,10 @@ subroutine hinit(iold,nend4,nend6,nscr)
             call error('hinit','bad LFC in mt=458.',' ')
          endif
       else
+         lfc=0
```
jlconlin: Why redefine `lfc` and `nply` here?

whaeck: Because that way, it will go through the thermal point evaluation branch of the heating calculation. The approximation made is based on thermal point.

PR #59, opened 2018-01-24
Feature/purr endf8

whaeck:
Some time ago, Paul Romano (ANL) proposed a fix to purr to correct a division by zero he observed for Po210 from ENDF/B-VIII.0b7. The fix by itself was straight forward and has been implemented in this pull request. However, when investigating the reason behind the division by zero, it become clear that the probability table was inherently wrong for Po210.

The way purr works is that it will generate a number of samples of cross section values corresponding to randomly sampled energy values over a predetermined energy range. This energy range is determined using (among other things) the lowest value of the average level spacing defined for the unresolved resonance parameters. It is here that things went wrong. To find this value, NJOY first initialises the dmin value to 500 after which it compares with the available data from the evaluation. If it is lower than dmin, NJOY sets the value of dmin to the new value until all values have been tested. For Po210, dmin remained 500 because the lowest value for the average level spacing was ~30000. This resulted in a negative value for the number of resonances nres and the emax value becoming negative while emin remained positive (these last two values determine the energy range).

This resulting energy value sampling led to the generation of cross section values of which 9500 were exactly the same. When constructing the probability bin limits, NJOY thus ran out of different cross section values to generate these bin limits. Due to the logic by which this is done in NJOY, it lead to the generation of bin limits that were actually smaller than the previous limit. As a result, no samples were accumulated in this bin and any following bins, giving a zero probability and zero cross section value. It is such a zero cross section value that caused the division by zero.

All this to say that by setting the initial dmin value to a higher value the problem will go away.

This problem does no occur for nuclides with an average level spacing of less than 500. For example, for U235, this is 0.7.

The following changes were implemented in this pull request:
- the original changes by Paul Romano
- setting dmin to 100000
- adding a test to issue an error message should this dmin value prove insufficient in the future
- adding a test to count the number of zero probability bins and occurrences of negative cross section values (NJOY issues a message in these cases)
- changes to the output results of test 21 (ace generation for ENDF/B-VIII.0b Fe58): Fe58 has a minimum level spacing of 6025 but it did not lead to a division by zero as with Po210

The version number change still needs to be pushed up, pending additional changes to be requested.

### DIFF
--- a/src/vers.f90
+++ b/src/vers.f90
@@ -3,7 +3,7 @@ module version
    ! These values are updated during the NJOY revision-control process.
    implicit none
    private
-   character(8),public::vers='2016.23'
-   character(8),public::vday='18Jan18'
+   character(8),public::vers='2016.24'
+   character(8),public::vday='24Jan18'
 end module version
 


src/purr.f90
```diff
@@ -2401,6 +2408,14 @@ subroutine unrest(bkg,sig0,sigf,tabl,tval,er,gnr,gfr,ggr,gxr,gt,&
            nmr(ixx),temp(itemp),(tabl(i,ixx+1,itemp),i=1,nbin)
       enddo
    enddo
+   if (izeroprob.gt.0) then
+      write(strng,'(''ptable has '',i3,'' zero probability bins'')') izeroprob
+      call mess('purr',strng,' ')
+   endif
+   if (ibadxs.gt.0) then
```
jlconlin: 👍

src/purr.f90
```diff
@@ -2401,6 +2408,14 @@ subroutine unrest(bkg,sig0,sigf,tabl,tval,er,gnr,gfr,ggr,gxr,gt,&
            nmr(ixx),temp(itemp),(tabl(i,ixx+1,itemp),i=1,nbin)
       enddo
    enddo
+   if (izeroprob.gt.0) then
```
jlconlin: 👍

jlconlin:
Do NOT merge this unless the test builds pass.

jlconlin:
This is still failing on test 24. Seems like something still isn't fixed.

whaeck:
Yes, but it is in acer, in the plotting part for 3D graphs - probably the
angular distributions or the spectra. It has nothing to do with the changes
I made in purr.

2018-01-24 12:54 GMT-07:00 Jeremy Lloyd Conlin <notifications@github.com>:

whaeck:
Changes were made to address the issue with test 24. The problem was due to the secondary neutron energy and angle distribution for n,4n in Pu239.

PR #64, opened 2018-01-30
Add more beta values for the calculations of S(ab.b) as free gas in THERMR

marquezj:
Extended the beta grid to work with higher incident energies.

### DIFF
--- a/src/thermr.f90
+++ b/src/thermr.f90
@@ -1793,7 +1793,7 @@ contains
    !--use free-gas s(alpha,beta).
   200 continue
    nalpha=1
-   nbeta=9
+   nbeta=30
    allocate(alpha(nalpha))
    allocate(beta(nbeta))
    allocate(sab(nalpha,nbeta))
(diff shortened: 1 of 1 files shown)


whaeck:
Hi Ignacio,

I actually put my comment in the issue, not here :-)

Thank you for sharing this, it does solve the issue we encountered. Dan Roubtsov also contributed a few changes in acer related to the thermal scattering so I would propose to include your changes in that branch (you can see it as the "feature/acer-endf8" branch).

I would also like to add a test that would allow us to signal this issue in the future. I was thinking about something like:
````
if ((de/bk/temp).gt.b(30)) call mess( ... )
````
in which de is some realistic upper limit value for the energy transfer.

Best regards,
Wim

whaeck:
I have reviewed test 1 results to assess the changes. We see differences for the highest two incident energies (1.16 and 1.28 eV) and the lower outgoing energies (below .5 eV) so the changes appear only for the larger values of the energy transfer (larger beta values).

The lower outgoing energies up to ~0.5 eV are different and the associated cosines are slightly different, but not by much. The cosines differ at the third or fourth digit after the decimal point (e.g. -0.990690 vs -0.990396). I do not think this is something to really worry about.

whaeck:
The pull request for the feature/acer-endf8 branch has been made and includes these changes. This pull request can thus be closed.

Thanks again for the help.

Best regards,
Wim

PR #67, opened 2018-02-06
Feature/acer endf8

whaeck:
Small fixes in this pull request:
- expanded the beta grid for free thermal gas scattering in thermr up to beta=3500 to properly process H1 at 0.1 eV
- fixed a NaN appearing in thermal scattering plots for H-ZrH, NJOY wanted to generate an empty plot

### DIFF
--- a/src/thermr.f90
+++ b/src/thermr.f90
@@ -1792,8 +1792,8 @@ contains
 
    !--use free-gas s(alpha,beta).
   200 continue
-   ! the beta grid is fixed to 30 beta values (previously only the first 9)
-   ! following an issue with Fe56 from ENDF/B-VIII.0b7 - a better solution
+   ! the beta grid is fixed to 45 beta values (previously only the first 9)
+   ! following an issue with Fe56 and H1 from ENDF/B-VIII.0 - a better solution
    ! would be to contruct the grid adaptatively based on emax and tempr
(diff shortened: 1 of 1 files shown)


src/aceth.f90
```diff
@@ -1554,40 +1554,42 @@ subroutine tplots(nout,hk)
...
-      do j=1,nbini
-         ep=xss(loc+1)
-         p=xss(loc+2)
-         if (p.lt.zmin) p=zmin
-         if (e.ge.ymin.and.e.le.ymax.and.&
-           ep.gt.xmin.and.ep.le.xmax) then
-            write(nout,'(1p,2e14.6,''/'')') ep,p
-         endif
-         loc=loc+nang+3
+   if (zmax.gt.zero) then
```
jlconlin: What is `zmax`? What if it isn't `>0`?

whaeck: zmin and zmax are the maximum values for the z axis in a 3D plot. Because NJOY does not know which values to use, it will go over the data and determine which values to use. As such, zmin is initialised to 1000 and zmax to 0.
So, if zmax is not larger than 0, then this means NJOY did not find any data that would go on the plot in the x,y range it was looking at. Therefore: just skip the plot, which is what the if statement does.

PR #78, opened 2018-04-06
Delete re-definition of cfrac as a local variable in SKOLD

marquezj:
The variable cfrac is defined twice in LEAPR: once as a global variable and again inside of the subroutine SKOLD. This change deletes this second definition.


whaeck:
Hi Ignacio,

Dan Roubtsov (@danrouzen) mentioned this one to me quite some time ago but I have not had the time to handle it. Thank you for doing this one. I am currently running our test suite on it (as a matter of procedure, I do not think this change will have an impact anyway).

I would like to ask you to change the NJOY version and date in the vers.f90 file as well. Could you set the version to 2016.33? We already have a fix set up with the number 2016.32 and we will probably merge that one into the master today.

Thanks again.

Best regards,
Wim

marquezj:
Done. 

Yes, this is something we have been discussing with Dan, and that little mistake made it impossible to run the new D2O inputs in LEAPR.

whaeck:
Good to know. Do you think this merits a test in the test suite? Apparently your D2O input goes into parts of the code not tested by the non regression tests, so I would actually encourage you to add one. These tests are only for coverage so they do not have to be exhaustive (only one or two temperatures, etc.).

marquezj:
I will discuss this with Dan. The thing is, the evaluations for light and heavy water were done with NJOY99, and NJOY2016 gives slightly different numbers (nothing dramatic so far, but they would fail a check based on diff). I think we should check if those differences are significant and then prepare a "canonical" run of H2O and D2O using NJOY2016 for a couple of temperatures as you say. That test would give the users the tranquility that a build of LEAPR (or, better, LEAPR+THERMR+ACER) reproduces well the new scattering kernels.

marquezj:
@whaeck I just added test 33: a simplified version (only one temperature) of the ENDF/B-VIII.0 LEAPR inputs for deuterium and oxygen bound in heavy water. To make them both run in the same input I had to add another change to LEAPR: at the beginning of ENDOUT, NSH must be set to zero, because the previous LEAPR run leaves it as one.

whaeck:
@marquezj Thanks, I was actually planning on pulling this one in without the test (knowing that we could add it later). I will look over it and proceed from there.

Thank you again for the help. We'll probably see each other at PHYSOR?

jlconlin:
So I've pulled from @marquezj branch and am running the tests right now. (Mac, gcc.) It seems that Test22 fails. Have you seen this failure?

marquezj:
Great.
Regarding PHYSOR, sadly I will not be there. Have a nice time in Cancún.

marquezj:
@jlconlin no, test22 passes in my test system (Linux, gcc).

marquezj:
@jlconlin can you upload the diff for that test?

whaeck:
The CMakeLists.txt file update makes it go away - the system got confused I think. Test 33 on the other hand fails with a floating point exception (Debug build):

[wim@silverton 33]$ ../../njoy < input 

 njoy 2016.33  09Apr18                                       04/16/18 09:28:39
 *****************************************************************************

 leapr...                                                                 0.0s

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7f4fcaa7126f in ???
#1  0x7f4fcbc0dcd3 in terps
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/leapr.f90:1104
#2  0x7f4fcbc1048a in trans
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/leapr.f90:905
#3  0x7f4fcbc19647 in __leapm_MOD_leapr
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/leapr.f90:370
#4  0x401d5e in njoy
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/main.f90:259
#5  0x401f02 in main
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/main.f90:112
Floating point exception

jlconlin:
I'm seeing the same error as @whaeck when building in debug mode. This error does *not* exist when building in release mode. The change to the CMakeLists.txt file fixes the error with Test22.

whaeck:
The problem is due to an integer overflow in the terps function. The division of be by delta is converted into an integer which becomes too big for a standard kind=4 integer. I tried changing the i variable to integer(kind=8) and use i=int(be/delta,8). It made the problem go away but it crashes later with another error:
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7fba2c50b26f in ???
#1  0x7fba2cd15373 in ???
#2  0x7fba2d50a434 in __endf_MOD_terp1
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/endf.f90:1631
#3  0x7fba2d694f34 in skold
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/leapr.f90:2816
#4  0x7fba2d6b3a9a in __leapm_MOD_leapr
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/leapr.f90:379
#5  0x401d5e in njoy
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/main.f90:259
#6  0x401f02 in main
	at /home/wim/Projects/NJOY2016/njoy2016-marquezj/src/main.f90:112
Floating point exception

marquezj:
Ok, it seems that the comparison in the next line deals with that overflow. Without the overflow check from the compiler, if be/delta is too big, i becomes negative and returns on line [1105](https://github.com/marquezj/NJOY2016/blob/bcb25937996a46aafb9ca858807a631bc01aa6ab/src/leapr.f90#L1105). I will try a check that is more robust and modern than that.

whaeck:
I just compared the output from the Debug run with the kind=8 correction with the Release output with and without the kind=8 correction. Up to the point where the Debug version fails due to the second floating point exception, all runs give the same results (up to line 108829 of the output file, after which the second floating point exception occurs). For some reason, the Release build seems to use 64 bit integers I think.

My correction for kind=8 in terps is as follows:
```

   real(kr) function terps(sd,nsd,delta,be)
   !--------------------------------------------------------------------
   ! Interpolate in a table of S(alpha,beta) for a required beta.
   ! Used in trans.
   !--------------------------------------------------------------------
   ! externals
   integer::nsd
   real(kr)::delta,be
   real(kr)::sd(nsd)
   ! internals
   integer(kind=8)::i
   real(kr)::bt,btp,st,stp,stt
   real(kr),parameter::slim=-225.e0_kr
   real(kr),parameter::zero=0

   terps=0
   i=int(be/delta,8)
```

I will dig a bit further to see where the second floating point exception cones from.

whaeck:
That second floating point happens in the terp1 interpolation routine from endf.f90. The input of the routine are the x1,y1 and x2,y2 points as well as the x value at which it must be interpolated and the y value in which to store the value.

The routine actually checks whether or not y1 is zero when using log-log (interpolation type=5) but it does not check if y2 is zero as well. The problem occurs when y1 is non zero (in this case 1e-17, so close to zero) and y2 is exactly zero. In the interpolation formula, we need to calculate log(y2/y1) or log(0) which does not compute :-)

The problem actually happens when the coherent scattering function you're trying to interpolate on is very small or zero altogether. As such, setting the interpolated value to zero by default actually makes sense.

By changing lines 2809-2811 in leapr.f90 to the following, the problem goes away and leapr runs to completion:
```
         if (ssm(i,kk-1,itemp).eq.zero.or.ssm(i,kk,itemp).eq.zero) then
            scoh(j)=zero
         else
            call terp1(alpha(kk-1),ssm(i,kk-1,itemp),&
              alpha(kk),ssm(i,kk,itemp),ap,scoh(j),5)
            scoh(j)=scoh(j)*sk
         endif
```

Comparing the release run results (prior to these changes) with the debug run with these changes reveals some differences: where the previous release run had very small values (1e-30 or lower), the debug run now displays 0.0.

whaeck:
I also compared the endf tapes being produced, I only see a difference for the highest alpha values with a non zero S(a,b) associated to them. The S(a,b) values are small (smaller than 1e-20) and the debug run gives me slightly different values. All other values are the same.

marquezj:
I added checks to avoid raising FP exceptions. Now all test pass in Debug mode. Now that I see your comment, the implementation is slightly different, but to the same effect.

whaeck:
I updated the reference tapes for test 33. The changes were within tolerance but I wanted the reference tapes to reflect the results for the current implementation versus the previous one. Test 22 and 23 are unaffected by these changes.

This is ready to be merged, unless you have something else lined up. If it is OK for you, I will go ahead and merge this in the master branch.

jlconlin:
Thanks for the good work to make this Pull Request a good one. It seemed simple on the surface, but had some problems hiding.

Issue #82, opened 2018-04-30
Out off bounds exception in purr


whaeck:
The attached input file crashes in purr due to an out of bounds exception on an array when writing to the output files in purr.

The original ENDF tape is ENDF/B-VII.1 U235.

[input.txt](https://github.com/njoy/NJOY2016/files/1961654/input.txt)

whaeck:
The problem occurs when purr copies the index of mf1 mt451 to the new output file (in which it needs to add mt152 and mt153 in mf2). It would appear that the c array containing the new index is not large enough (by 6 elements). Why this does not pop up in other cases is beyond me.

whaeck:
Will be fixed in NJOY 2016.35, see feature/purr-endf8 branch. This branch is still under development but it will be pulled into the master branch soon (when I finish adding tests for the ptable flag update).

whaeck:
This has been addressed in pull request #83 and should be available in a couple of days in the master branch.

PR #83, opened 2018-05-01
Feature/purr endf8

whaeck:
This is the latest pull request related to updates in purr. This pull request handles a minor issue (#82) and a more complex problem that appeared when trying to handle the Pu240 issue which was already the subject of a fix in pull request #73.

Pull request #73 fixed the problem when the total cross section in the unresolved resonance region is smaller than the sum of its components. However, the Pu240 total cross section was also larger than the sum of its components in the upper energy range of the unresolved resonance region. This problem was not fixed in pull request #73 so the probability table for Pu240 is still only partly correct. The backgrounds for LSSF=1 should be set to zero if the total cross section is different from the sum of its components. Because the probability table only handles 4 reactions (total, elastic, fission and capture), it is still possible to have a total cross section larger than the sum of its components above the competition threshold if there is competition reaction (inelastic and/or absorption). 

However, it became clear that NJOY only tested competition up to MT102 (neutron capture) and explicitly omitted checking reactions like MT5 (n,anything), MT103 (n,p) or MT107 (n,a). As a result, for some nuclides NJOY did not capture the competition flags properly, leading MCNP or other codes using the probability tables to potentially calculate the total cross section in the unresolved resonance region incorrectly. In most cases, the impact of this error is assumed to be small as the reactions with an MT number above MT102 are generally quite small. However, there may be nuclides for which this is not true.

The following changes were made to purr to properly handle these issues:
- The LSSF=1 sanity checks for the backgrounds are now run in the function that reads the unresolved resonance parameters and the background cross sections in MF3.
- The total background for LSSF=1 is set to zero in the following cases: the total cross section is smaller than the sum of its components, the total cross section is larger than the sum of its components when there is no competition and the total cross section is larger than the sum of its components below the competition threshold when there is competition. Only when the relative difference between the total and the sum of its components is greater than 1e-6, purr will signal this as a message (the total background is still set to 0)
- The detection of competition has been rewritten to expand the range beyond MT102 and to include MT5 as well.
- The inelastic and absorption competition flags are now handled the same in purr and acer. Previously, purr compacted these flags into a single flag which acer than unpacked for use in the ace files. Now purr puts both flags in MF2 MT153 in the output PENDF file as used by acer (-1 meaning no competition, 0 meaning MCNP has to figure it out by itself or a positive non zero value indicating the MT number of the single competitive reaction).

Tests 35 to 42 were added to test the inelastic and absorption competition flags determined by `PURR` and used in `ACER` to be included in ACE files. Previous versions of NJOY 2016 (prior to 2016.30) incorrectly handled these flags. NJOY only tested competition up to MT102 (neutron capture) and explicitly omitted checking reactions like MT103 (n,p) or MT107 (n,a). As a result, for some nuclides NJOY did not capture the competition flags properly, leading MCNP or other codes using probability tables to incorrectly calculate the total cross section in the unresolved resonance region.

Tests 10, 19, 21 and 31 were updated following these changes. The probability tables only changed for test 31 (the Pu240 problem previously described). Tests 35 to 42 were added to cover all possible combinations of the competition flags (no competition, only elastic competition, only absorption competition and both) for each possible value of the LSSF flag (0 or 1). Some of these tests may overlap with the previously mentioned tests 10, 19, 21 and 31 but, frankly my dear, I don't give a damn ;-)

### DIFF
--- a/tests/33/input
+++ b/tests/33/input
@@ -1,6 +1,6 @@
 leapr
 24  /NOUT
-D in D2O, CAB Model from molecular dynamics calculations / TITLE
+'D in D2O, CAB Model from molecular dynamics calculations' / TITLE
 1 2 200  / NTEMPR IPRINT IPHON
 11 1002 0 0 2E-38 / MAT ZA ISABT ILOG SMIN
 1.996800 3.395 2 0 0 2  / AWR SPR NPR IEL NCOLD NSK
@@ -362,7 +362,7 @@ D in D2O, CAB Model from molecular dynamics calculations / TITLE
 / end leapr
 leapr
 34  /NOUT
-O in D2O, CAB Model from molecular dynamics calculations/ TITLE
+'O in D2O, CAB Model from molecular dynamics calculations'/ TITLE
 1 2 200  / NTEMPR IPRINT IPHON
 51 8016 0 0 2E-38 / MAT ZA ISABT ILOG
 15.857511 3.7939 1 0 0 2  / AWR SPR NPR IEL NCOLD NSK


whaeck:
@paulromano 
This will depend on how large the competitive cross sections are. For instance, if we are talking about a total cross section of 10 barn and the competition is only 1mbarn or microbarn, then the impact will be negligible.

To be clear: only the flags have changed, The actual probability tables have not (as long as your evaluation did not suffer from a problem similar to the one of Pu240). So, if a code ignores the flags and simply recalculates the total from the available cross sections in the ptable and the other reactions, then you should NOT see a change in the results. If you rely on the flags, that may be a different matter. For example: ENDF/B-VIII.0 U235 now has a competition flag for absorption set to 0 while with NJOY 2016.29 or lower it was set to -1.

We are using this version to regenerate the ENDF/B-VIII.0 ace files. @jlconlin will run the ICSBEP benchmarks on the new files so that we can compare to the old version to see some of the impact.

I will however study this in more detail, e.g. using transmission of neutrons through a plate at energies corresponding to the unresolved resonance region and using some criticality benchmarks. If you have an idea of some crits that would be impacted, I would like to get your list of benchmarks.

kahlerac:
A cynic will say that anything dealing with the URR is wrong in detail but hopefully close to correct on average.  I believe the code changes here are valid and represent a move toward being more correct on average ... but

and there's always a "but".  This update is changing the mf2/mt153 format and there is an implicit assumption in the acefc coding that it will never process a pendf tape from 2016.34 or earlier (i.e., a tape with the original mf2/mt153 format).  Maybe there should be coding in acefc to handle both new and old mf2/mt123 formats.  The "new" format gives us iinel and iabso from a CONT record and sets the l2h term in the subsequent LIST to zero.  The "old" format doesn't care about the CONT record but extracts iinel and iabso from the LIST (non-zero) l2h term.  This seems different enough that coding can be developed to handle either situation.  Thoughts?

Issue #97, opened 2018-08-20
Changing ngmax leads to problem in GAMINR


whaeck:
When using more then 400 groups in the user defined group structure in GAMINR, GAMINR will instruct the user to increase ngmax to a higher value. The default value is set to 400.

When doing this, GAMINR will however try to read the group structure into arrays that are still limited to the 400 element size (the size of the arrays was hardcoded to be equal to 400 instead of using the ngmax parameter), causing the issue.

whaeck:
Fixed in #98 - to be merged soon, no need to keep this open.

PR #98, opened 2018-08-20
Fixed issue #97

whaeck:
Fix for issue #97


whaeck:
The failing travis tests are due to time outs.

whaeck:
One of the tests has been "pending" for quite some time now. When checking the output on Jenkins, all NJOY tests pass but something went wrong while exiting afterwards, which appears to generate the "pending" test.

PR #99, opened 2018-09-10
Feature/updated constants

jlconlin:
In this Pull Request, I've updated the physical constants to something a bit more modern. Of course, this changes virtually all of the reference tapes. Exactly what to do with this is a matter of discussion.

### DIFF
--- a/src/samm.f90
+++ b/src/samm.f90
@@ -1295,27 +1295,27 @@ contains
          kza(i,ier)=0
          lpent(i,ier)=1
       else if (mt(i,ier).eq.103) then  ! proton
-         ema(i,ier)=amassp/amassn
+         ema(i,ier)=pnratio
          spina(i,ier)=0.5e0_kr
          kza(i,ier)=1
          lpent(i,ier)=1
       else if (mt(i,ier).eq.104) then  ! deuteron
-         ema(i,ier)=amassd/amassn
+         ema(i,ier)=dnratio
          spina(i,ier)=1
          kza(i,ier)=1
          lpent(i,ier)=1
       else if (mt(i,ier).eq.105) then  ! triton
-         ema(i,ier)=amasst/amassn
+         ema(i,ier)=tnratio
          spina(i,ier)=0.5e0_kr
          kza(i,ier)=1
          lpent(i,ier)=1
       else if (mt(i,ier).eq.106) then  ! he3
-         ema(i,ier)=amassh/amassn
+         ema(i,ier)=hnratio
          spina(i,ier)=0.5e0_kr
          kza(i,ier)=2
          lpent(i,ier)=1
       else if (mt(i,ier).eq.107) then  ! alpha
-         ema(i,ier)=amassa/amassn
+         ema(i,ier)=anratio
          spina(i,ier)=0
          pa(i,ier)=0
          kza(i,ier)=2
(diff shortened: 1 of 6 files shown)


jlconlin:
Note: @kahlerac did most of the work on this and gave me the upn file before he retired. I incorporated the changes into git and had to make a few minor changes to get it to compile.

whaeck:
So I looked at the changes in the output files. Can you rerun the tests using a tolerance of 1e-5? That'll allow us to quantify the overall impact of changing the values of the physical constants.

jlconlin:
I tested with 1E-5 relative tolerance. This didn't change the test results. 

When the comparison is done, the first thing that is done is to count the number of lines in the the generated file and the reference file. If the number of lines are not the same, then no other checking is done. In fact, additional checking can't be done since we do everything line-by-line.

jlconlin:
Now, perhaps we should change the way that we compare references files. There is an argument to be made there.

whaeck:
Well, we do test non regression, so if the number of lines is different there is definitely something going on. In those cases a by hand verification is warranted.

I assume that in this case, no files actually get additional lines?

jlconlin:
No, there are plenty that get additional lines.

jlconlin:
I've updated the constants even more. I double-checked the values put in previously. I used the constants in the ENDF manual and from NIST (when not in the ENDF manual).

src/samm.f90
```diff
@@ -1300,27 +1295,27 @@ subroutine ppdefs(ier)
...
          lpent(i,ier)=1
       else if (mt(i,ier).eq.106) then  ! he3
-         ema(i,ier)=he3/neutron
+         ema(i,ier)=amassh/amassn
          spina(i,ier)=0.5e0_kr
          kza(i,ier)=2
          lpent(i,ier)=1
       else if (mt(i,ier).eq.107) then  ! alpha
-         ema(i,ier)=alpha/neutron
+         ema(i,ier)=amassa/amassn
```
whaeck: See above: these values return a couple of times. Maybe introduce the following parameters:
pnratio=amassp/amassn
dnratio=amassd/amassn
etc.

src/thermr.f90
```diff
@@ -2300,7 +2300,7 @@ subroutine calcem(temp,itemp,iold,inew,ne,nex)
       write(nsyso,'(/''            mu            theta      dsigma/dmu'')')
       do i=1,nmu
          write(nsyso,'(i5,1x,f15.8,1x,f12.4,1x,1p,e14.7)') i,uj(i),&
-               acos(uj(i))*180.0/3.14159265359,sj(i)/2.0
+           acos(uj(i))*180/pi,sj(i)/2
```
whaeck: This was the one that caused a lot of issues in the thermr outputs? Try acos(uj(i))*180.0/pi,sj(i)/2.0 instead to see if it continues.

whaeck:
By the way, do not forget to update the version number and the date.

src/samm.f90
jlconlin: Yeah, but where else are they used? Are they used outside of `samm.f90`? If not, then it shouldn't be part of `phys.f90`. I also don't want to make this a big refactoring of of the entire codebase. We have to draw the line somewhere.

whaeck: groupr too.

src/leapr.f90
```diff
@@ -1959,14 +1957,19 @@ subroutine coldh(itemp,temp)
...
-   if (law.gt.3) sampi=sampid
+   if (law.gt.3) then
+     amassm=6.69E-24_kr
+     ! amassm=2*(amassd+amasse)*amu*ev/(clight*clight)
+     sampc=sampcd
+     bp=hbar/2*sqrt(2/ded/ev/dmass)/angst
+     sampi=sampid
+   else
+     amassm=3.3464e-24_kr
+     ! amassm=2*(amassp+amasse)*amu*ev/(clight*clight)
```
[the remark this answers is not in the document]
jlconlin: When I change this to use the formula, I get a different value for `amassm`, different by 12 orders of magnitude! So I don't think that formula is correct. I chose to leave it as is until we know what is going on.

PR #101, opened 2018-09-13
Feature/fix

whaeck:
Changes to gaspr to solve issue #100.

The levels are now systematically ignored because reconr will always add the summation cross sections.


jlconlin:
Thanks for your approval @kahlerac.

Issue #106, opened 2018-10-10
Typo in hbar


whaeck:
After refactoring the constants into the phys.f90 file, hbar (in erg) is defined as:
   real(kr),parameter,public::hbar=6.58219514e-16_kr*ev !Planck/2pi, erg

The numerical value has a missing digit in it, it should be:
   real(kr),parameter,public::hbar=6.582119514e-16_kr*ev !Planck/2pi, erg

E.g. see ENDF manual appendix H.

Be aware that this minor difference has an impact on the 4th or 5th significant digit of all cross section values coming out of NJOY and has been present in the code since 27 days. This will be fixed as soon as possible.

whaeck:
Fixed in #107

whaeck:
Additional typos were discovered for the electron and He3 (helion) mass values, reported by @kahlerac. These have been fixed in bugfix/constants.

whaeck:
Fixed additional issues in #115

PR #115, opened 2018-11-19
Bugfix/constants

whaeck:
Additional fixes to the physical constants: electron and helion mass. See #106 

No impact on test results (no changes, these constants do not seem to be used in any of the tests we currently run).

### DIFF
--- a/src/phys.f90
+++ b/src/phys.f90
@@ -37,7 +37,7 @@ module physics
    real(kr),parameter,public::amassp=1.007276466879e0_kr !proton
    real(kr),parameter,public::amassd=2.013553212745e0_kr !deuteron
    real(kr),parameter,public::amasst=3.01550071632e0_kr   !triton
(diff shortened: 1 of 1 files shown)


src/phys.f90
```diff
@@ -37,9 +37,9 @@ module physics
    real(kr),parameter,public::amassp=1.007276466879e0_kr !proton
    real(kr),parameter,public::amassd=2.013553212745e0_kr !deuteron
    real(kr),parameter,public::amasst=3.01550071632e0_kr   !triton
-   real(kr),parameter,public::amassh=3.014932234673e0_kr !hellion (3)
+   real(kr),parameter,public::amassh=3.01493224673e0_kr !hellion (3)
```
jlconlin: While we're at it, we should fix the typo in the comment. It should be `helion`, not `hellion`. See [Wikipedia](https://en.wikipedia.org/wiki/Helion_(chemistry))

whaeck: Done ;-)

Issue #116, opened 2018-12-17
GROUPR has problems due to NaNs coming from UNRESR


whaeck:
Thomas Saller (LANL) indicated an issue with ENDF/B-VII.1 Na22 (the same evaluation was also carried over into ENDF/B-VIII.0) leading GROUPR to produce NaN cross section values. It would appear that this is independent of the sigma0 values used (both 1e+10 and 0.1 barn produce some NaN values).

An input file to illustrate the issue is attached: [na22.txt](https://github.com/njoy/NJOY2016/files/2688205/na22.txt)

Now, ENDF/B-VII.1 Na22 is a file taken over from JEF 2.2, and it has a history of issues in the unresolved resonance region (PURR for instance also generates negative or zero cross section values in the unresolved resonance region) and it would appear that the problem is related to this as well.

In the calculation method used in UNRESR, a modified cross section sigm is calculated as the sum of the background cross section, the potential scattering cross section, the interference cross section and the sigma0 value. If the background is negative (which is the case for Na22, sigb=-4.77 at 15000 eV) and the sum of the potential scattering and sigma0 value is smaller in absolute value compared to that backgound (sigma0=0.1, spot=4.05 and sint=-0.21  at 15000 eV), then the modified sigm is negative. In the example at 15 keV, sigm=-0.82.

At some point in the calculation, UNRESR will need to calculate a value beta defined as sigm divided by a factor (which is a function of the unresolved resonance parameters). At 15 keV for sigma0=0.1, beta=-1.4e-2 This value is then used later on in a square root of ( 1 + beta ) / beta. As a result, if beta is negative and the absolute value is larger than 1, than the minus sign of beta cancels out int he calculation of the square root. However, is beta is negative and the absolute value is smaller than 1, then we are taking the square root of a negative value. In the example at 15 keV, this leads to a NaN value which then propagates through to the cross section value.

In order to better diagnose this problem in the future, a message should be added to UNRESR to indicate this issue, as we have already done in PURR (where we signal negative cross sections and zero probabilities, or issues with the total cross section not summing to the sum of its components).

As a workaround, I would suggest using a different evaluation.

whaeck:
Fixed with #117

PR #117, opened 2018-12-18
Fix/unresr

whaeck:
Additional messages in UNRESR for signaling issue #116.

The changes are as follows:
- a message is issued when the background plus the potential scattering, the interference and sigma0 value are negative
- a message is issued when the beta value is within the -1 to 0 range (which causes a sqrt to return NaN) 

Questions:
- the first message could be slightly changed so that it is issued when the background plus the potential scattering and the interference are negative, regardless of the value of sigma0. In my opinion, the fact that this value is negative before adding sigma0 is an indication that the evaluation is no good whatever happens.
- the second message could be transformed into an error and that would stop the processing dead in its tracks (which degrades the behaviour from before these changes - it was silently ignored). In my opinion, the results will be inherently wrong, which would merit the use of error instead of mess.

### DIFF
--- a/src/unresr.f90
+++ b/src/unresr.f90
@@ -944,10 +944,10 @@ contains
    sigbt=sigbkg(1)+spot+sint
    do is=1,nsig0
       sigm(is)=sigbt+sig0(is)
-	  if (sigm(is).lt.zero) then
+      if (sigm(is).lt.zero) then
          call mess('unresl', 'Negative background xs in urr may cause issues',&
-		                     &'Check the evaluation')
-	  endif
+                         &'Check the evaluation')
+      endif
    enddo
    ispot=1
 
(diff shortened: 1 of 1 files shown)


src/unresr.f90
```diff
@@ -1103,6 +1110,11 @@ subroutine unresl(ee,tt,sig0,nsig0,sigbkg,sigu,arry)
                   sti=gg(5)/del(itp)
                   do is0=1,nsig0
                      beta=sigm(is0)/s0u
+					 if (beta.lt.zero.and.abs(beta).lt.one.and.noissue) then
```
jlconlin: This is some strange indenting. Can you fix it?

src/unresr.f90
```diff
@@ -1103,6 +1110,11 @@ subroutine unresl(ee,tt,sig0,nsig0,sigbkg,sigu,arry)
...
+					    call mess('unresl','Square root of negative number detected',&
```
jlconlin: I like the idea of making this an error instead of a message. **Unless** this causes too many evaluations to fail. That will really get us into trouble if suddenly we can't process our data. @kahlerac can you weigh in on this?

src/unresr.f90
whaeck: That appear to be tabs inserted by notepad++, I'll fix that

jlconlin: Get yourself a [real editor](https://www.vim.org) 😄

src/unresr.f90
kahlerac: All,
   I think an "abort" is the correct action to take, and in fact was one of the first updates I made shortly after arriving at LANL in the Fall 2006 ... see update 167 to njoy99.  But our European friends didn't like it and asked that it be reverted to a message, which was done in update 265.  Guess it can't hurt to try again!

Issue #118, opened 2019-01-23
GROUPR only produces infinite dilute fission matrix (MF6 MT18) data


whaeck:
Thomas Saller (LANL) indicated an issue when processing ENDF/B-VIII.0 Pu236 multigroup data into the NDI format when trying to create NDI files for various sigma0 values (up to now, infinite dilute data was the norm). During the formatting process, the formatting tool indicated an issue with MF6 MT18 in the GENDF file that GROUPR produced.

An input file to illustrate the issue is attached: [pu236.txt](https://github.com/njoy/NJOY2016/files/2789272/pu236.txt)

This input file uses three sigma0 values: infinite dilute, 1e+5 and 1e+3. The "usual suspects" (total, elastic, fission and capture) have multigroup cross sections as a function of sigma0 value, as expected. However, only elastic scattering appears to have a scattering matrix as a function of sigma0 while only the infinite dilute fission matrix is given in the GENDF file.

NOTE: all nuclides with fission have this issue.

whaeck:
Branch fix/groupr contains changes to GROUPR to solve this problem.

The use of only an infinite dilute fission matrix/vector was due to this being hardcoded in the init subroutine which sets the size for the ans(il,iz,ig) array (il and iz represent the legendre order and the sigma0 values). Previously, for the fission matrix, the number of sigma0 values was hardcoded to be equal to 1. This has been changed to be equal to the number of sigma zero values given by the user. The legendre order is always 0 (isotropic), this has not been changed. 

By introducing a few additional loops over il and iz in the appropriate places instead of using ans(1,1,ig), the GENDF file now gives us the data as a function of sigma zero as well.

The ERRORR module was checked as well, as the GENDF file is used as input for the covariance processing. No changes appear to be required. The routine that reads the GENDF file uses a general approach (independent of which section is to be read) so the changes to the format should be transparent for ERRORR (please note that ERRORR only uses infinite dilute data from the GENDF file). The existing tests using ERRORR and COVR are not impacted by the changes (the input files in question only use a single sigma0 value). A test should be added using ERRORR alongside a GROUPR run that uses more than 1 sigma0 value.

jlconlin:
Good sleuthing to uncover the basis of this issue!

whaeck:
Some of the data MF6 MT18 written to the NJOY output file still needs to be changed (in some cases, GROUPR still writes only the infinite dilute data to the NJOY output file). The GENDF file however is fully fixed.

[...]

PR #119, opened 2019-02-14
Fix/groupr

whaeck:
This pull request addresses issue #118 

GROUPR was modified to produce MF6 MT18 data for each sigma0 value. A test was added to verify that this does not impact ERRORR processing. Other tests were modified to also provide the GROUPR GENDF file in ascii for testing this new feature.

### DIFF
--- a/src/vers.f90
+++ b/src/vers.f90
@@ -3,7 +3,7 @@ module version
    ! These values are updated during the NJOY revision-control process.
    implicit none
    private
-   character(8),public::vers='2016.48'
-   character(8),public::vday='18Dec18'
+   character(8),public::vers='2016.49'
+   character(8),public::vday='25Jan19'
 end module version
 


src/groupr.f90
```diff
@@ -706,7 +706,9 @@ subroutine groupr
   502 continue
    if (mtd.eq.455) then
       do i=1,ndelg
-         ans(i,1,1)=dntc(i)
+         do iz=1,nz
+            ans(i,iz,1)=dntc(i)
```
jlconlin: If I understand this correctly, this loop fills in the `ans` array for multiple columns instead of just the first one, right?

src/groupr.f90
```diff
@@ -716,7 +718,9 @@ subroutine groupr
       do i=1,ng2
          j=i
          if (mtd.eq.455) j=i+1
-         ans(il,1,j)=ff(il,i)
+         do iz=1,nz
+            ans(il,iz,j)=ff(il,i)
+         enddo
```
jlconlin: Now this loop fills in the last dimension of the `ans` array.

src/groupr.f90
whaeck: Yes. The size of the ans array is set in the init routine. Before the change, GROUPR only assigned the data for the iz=1, the first sigma0 value (by default infinity). Because the MF6 MT18 data was made dependent on the number of sigma0 values, the spectrum (this is what mt455 refers to) needs to be assigned or each sigma0 value as well. Otherwise we get zeros inside the GENDF file.

src/groupr.f90
whaeck: Indeed. The first value of the third dimension is never filled here if we're dealing with MT455. When we're processing an MT455 piece, we skip the first value of the third dimension (that's what the if statement before the do loop does) because we have already assigned to it.

PR #127, opened 2019-04-11
Fix/acer iza

whaeck:
Proposed fix for issue #126, more information can be found in the issue description.

The meaning of awn still needs to be evaluated.

### DIFF
--- a/src/acer.f90
+++ b/src/acer.f90
@@ -389,8 +389,8 @@ contains
          call error('acer','at least one za value must be given.',' ')
       endif
       do i=1,nza
-         if (izn(i).lt.1001) then
-            call error('acer','a za value must be larger then 1000.',' ')
+         if (izn(i).le.zero) then
+            call error('acer','found invalid za numbers in izn.',' ')
          endif
       enddo
       mti=0


src/acer.f90
```diff
@@ -174,10 +174,9 @@ subroutine acer
    !    matd     material to be processed
    !    tempd    temperature desired (kelvin) (default=300)
    !    tname    thermal zaid name ( 6 char max, def=za)
+   !    nza      number of moderator component za values (default=3, max=16)
```
jlconlin: The default should be 1 since traditional/default card 8a requires only 1.

Thoughts?

src/acer.f90
```diff
@@ -360,24 +358,36 @@ subroutine acer
    else if (iopt.eq.2) then
       tempd=300
       tscr=' '
-      read(nsysi,*) matd,tempd,tscr
+      nza=3
```
jlconlin: Whatever we decide the default should be, it should be reflected here.

whaeck:
@jlconlin @dkentp 
Ready for review

whaeck:
@dkentp To clarify: whenever we do something like:
```fortran
read(nsysi,*) (izn(i),i=1,nza)
```
the program will try to read as many values as it can, up until encountering '/' or when the total number of values has been read (potentially going onto the next line as necessary). Adding nza as an input parameter could alleviate some issues due to this but I don't think there will be major concerns (in most cases, users already set / at the end of that line to not have to write values for the second and/or third iza).

src/acer.f90
```diff
@@ -360,24 +358,36 @@ subroutine acer
...
-      iza02=0
-      iza03=0
-      read(nsysi,*) iza01,iza02,iza03
+      do i=1,nza
+         izn(i)=0
+      enddo
+      read(nsysi,*) (izn(i),i=1,nza)
+      do i=1,nza
+         if (izn(i).eq.zero) then
+            nza=i-1
```
jlconlin: What happens if Card 8a looks like this:
```
40090 40091 40092 0 0 40096 /
```
That would still be valid input, but now you will have trouble when trying to print out the values. I believe this will still create a valid ACE file, but the `output` file will have some missing information. Perhaps the answer is that we only read until we find the first 0 value (or get to the end of the card.) I recognize this is a corner case, but something to think about.

tests/49/input
```diff
@@ -0,0 +1,33 @@
...
+    -21 -22 -23
+    4025 1/
+    .001/
+    296 /
+    0/
+heatr
+    -21 -23 -24/
+    4025 4/
+    302 402 443 444 /
+thermr
```
jlconlin: So we can simplfy (i.e., speed up) this test if we only use ACER, which is all we are really testing at this point. We can start with `tape31`.

src/acer.f90
```diff
@@ -389,8 +389,8 @@ subroutine acer
          call error('acer','at least one za value must be given.',' ')
       endif
       do i=1,nza
-         if (izn(i).lt.1001) then
-            call error('acer','a za value must be larger then 1000.',' ')
+         if (izn(i).le.zero) then
+            call error('acer','found invalid za numbers in izn.',' ')
```
jlconlin: 👍

PR #129, opened 2019-09-19
Update leapr.f90

vedantkm:
Changes allow leapr module to account for non-integer principal scattering atoms. Examples: H in YH1.9, H in ZrH1.6.

### DIFF
--- a/src/leapr.f90
+++ b/src/leapr.f90
@@ -16,7 +16,7 @@ module leapm
    real(kr)::za
    real(kr)::awr
    real(kr)::spr
-   integer::npr
+   real(kr)::npr
    integer::iel
    integer::nss
    integer::ncold,nsk
@@ -267,7 +267,7 @@ contains
    write(nsyso,'(/&
      &  '' awr for principal scatterer .......... '',f10.3/&
      &  '' free xsec for principal scatterer .... '',f10.3/&
-     &  '' number of principal atoms ............ '',i10/&
+     &  '' number of principal atoms ............ '',f10.3/&
      &  '' elastic option ....................... '',i10/&
      &  '' cold moderator option ................ '',i10/&
      &  '' s(kappa) option ...................... '',i10)')&
@@ -2456,8 +2456,8 @@ contains
    use physics ! provides pi,hbar,ev,amu,amassn
    use util    ! provides timer,error
    ! externals
-   integer::lat,natom,nbe,maxb
-   real(kr)::b(maxb),emax
+   integer::lat,nbe,maxb
+   real(kr)::b(maxb),emax,natom
    ! internals
    integer::i,j,k,imax,jmin,idone,ifl,i1m,nw
    integer::i1,i2,i3,l1,l2,l3,i2m,i3m


jlconlin:
@vedantkm This is a pretty straightforward Pull Request, nothing drastic here. I'm curious if this has any effect on the output? If we are not changing/improving things, why do this? Can you please share what you expect out of this? (I know we have talked, but others have not had the same conversations we have had.)

Also, please update the version number and date in the `vers.f90` file. This will be required before we can merge this in.

vedantkm:
Hi Jeremy, yes exactly. There were couple very minor changes in the code that allows me to use non-integers for npr card. 

The output for the integers will not be changed as the leapr module will still be accurate as integers are subset of real numbers. 

The key advantage comes as now we can produce Thermal Scattering Laws for H in ZrH2-x, H in YH2-x, and other noninteger elements in a compound materials. The specified materials here are crucial for the new era of microreactors. The phonon DOS in leapr module is normalized. So, in a sense, leapr cares about how many H atoms are occupied in a unit volume and how they vibrate. Integers are okay, however they cause an issue when you have defects in materials. From atomistic perspective, I can model these defects, and find out what my phonon DOS are for a given defect in the material (say YH1.9). I can use this normalized phonon DOS (for YH1.9). But I cannot use npr=2 (non-updated leapr) anymore as in realitynpr=1.9 as that's the "amount of H available in a unit volume". 

I believe this simple update allows users to create TSLs for atom vacancies in a material.

jlconlin:
@marquezj Can you comment on this? You have thermal scattering expertise that I don't have. I'm not sure what the implications are here.

marquezj:
Well, the way this was usually solved was by doing independent models for the metal and hydrogen atoms in the hydride, each one with npr = 1. The fractional stoichiometry is included later in the number density, when the library is used. This does not take into account the correlation between atoms, but that is not a major problem because H-1 is the dominant scatterer and it is mostly incoherent. Considering this, the current implementation is not a restriction to what @vedantkm wants to solve.

Now, a mixture of a non-integer natom with nss=1 / mss>0 could be used for non-stoichiometric coherent scatterers (something like BeO, but with a ratio different to 1 to 1). This is the only case where I think a fractional npr would make sense, but the treatment of coherent scattering needs reworking anyway.

¿@vedantkm: can you provide an example of using fractional npr?

vedantkm:
@marquezj right, but I would like to implement all that together. I am getting phonons of the metal-hydride system with defects (through atomistics), then use the correct NPR initially, a non-stoichiometric system would have an non-integer which is now already accounted. Basically, we want S(a,b) to account for the TSLs as we go farther from stoichiometry. What changes these non-stoichiometries bring to the TSLs is unknown but at least for now if we use correct NPR for H in YH1.65 v/s YH1.25, it wouldn't account for any additional uncertainties. 

I believe I added a correction to the coherent part as well. 

See attached paper (third page - single paragraph) for more details and how this implementation could have helped there. 

[ANS_Final_Draft.pdf](https://github.com/njoy/NJOY2016/files/3649120/ANS_Final_Draft.pdf)

marquezj:
@vedantkm I understand, the vacancies in the structure introduce modifications in the spectra, but this can be taken into account in separate runs for the metal and the hydrogen (as it is now). And, if you produce one library for YH_{1.25} instead of one for Y(YH_{1.25}) and one for H(YH_{1.25}), how would you assign the library to one nuclide when you process it later with THERMR?

vedantkm:
Hi @marquezj I am a bit confused. I am producing HinYH1.25 and YinYH1.25 with the leapr module and then passing it through THERMR. They are separate runs for H and Y. 

Let me clarify if I made it confusing. The phonons are produced together for YH1.25 system from which you retrieve the phonons for Y and H separately for YH1.25 system. 

Now leapr part. Y in YH1.25 leapr system is fine as it is because npr = 1. But for H in YH1.25 (a different run), I would want to use npr=1.25, correct? They both go through THERMR separately afterwards. 

Are you saying that regardless of the H stoichiometry, I need to use NPR = 1 for H in YH1.25? Since the mat card in mcnp will use Y 1, H 1.25?

marquezj:
Exactly.

The DFT simulation for Y(YH_{1.25}) will yield a slightly different spectrum than the simulation for Y(YH_{2}), but the spectra for Y and H will be separately normalized to 1. For a single atom run, npr is just a factor that is multiplied in LEAPR and divided in THERMR.

vedantkm:
Just to confirm, we would still need NPR=1.25 for H in YH1.25, for LEAPR. Would you agree?

marquezj:
You can use `npr=1.25` in LEAPR and `natom=1.25` in THERMR, but you will get **exactly** the same results as using `npr=1` and `natom=1`. It is just a normalization.

My point is: the change in LEAPR is not bad (although it should be accompanied by the corresponding change in THERMR), but it is unnecessary.

PR #133, opened 2019-09-26
Undoing changes that were not supposed to go into master branch (yet).

jlconlin:
Got a little hasty this morning.


whaeck:
Weren't there more change to gaspr? I'll have to check the original file ...

Issue #138, opened 2019-10-31
Charged particles with a LAW=5 and LIDP=1, because -1**0 = -1 and (-1)**0 = 1


whaeck:
A problem was detected when running acer to produce an incident charged particle ace file: NaN values were appearing in the ACE file for a specific case where the incident charged particle and the target where the same type, with a spin s = 0. I'll add an example input file and corresponding example evaluation later.

For charged particle elastic scattering in the ENDF file, we can use MF6 LAW=5. Whenever we do this for identical particles (LIDP=1) such as proton on proton, deuteron on deuteron, triton on triton, etc., we need to calculate the coulomb scattering cross section using formula 6.10:
```
\sigma_{ci} = \frac{2\eta^2}{k^2 (1-\mu^2)} \left[ \frac{1+\mu^2}{1-\mu^2} + \frac{(-1)^{2s}}{2s+1} cos( \eta \ln( \frac{1 + \mu}{1 - \mu} ) ) \right]
```
In the case where mu = 0, this reduces to:
```
\sigma_{ci} ~ \left[ 1 + \frac{(-1)^{2s}}{2s+1} \right]
```
or when the spin s = 0 as well:
```
\sigma_{ci} ~ 1 + (-1)^0 ~ 2
```

In acer, this formula is implemented in acefc.f90, on line 6428:
```fortran
               if (lidp.eq.1) sigc=((2*eta**2/wn**2)&
                 /(1-amuu**2))*((1+amuu**2)/(1-amuu**2)&
                 +(-1**i2s)*cos(eta*log((1+amuu)/(1-amuu)))&
                 /(2*spi+1))
```
For mu = 0, this becomes equivalent to:
```
\sigma_{ci} ~ \left[ 1 + \frac{-1^{2s}}{2s+1} \right]
```
or when the spin s = 0 as well:
```
\sigma_{ci} ~ 1 + -1^0 = 0
```
Later on (line 6439), this value would be used as the denominator in a division, thus leading to the NaN values in the resulting ACE file. Adding the missing parentheses in the coding solves the issue.

Half integer spins do not have the issue. For integer spins different from 0 (s=1,2, etc.) there still is the same issue, but it will not lead to NaN values (because of the 2s + 1 term in the denominator).

whaeck:
The issue appears in GROUPR as well. The branch fix/cpt fixes this both in ACER and GROUPR.

whaeck:
[...]

PR #139, opened 2019-11-01
Fix/cpt

whaeck:
A number of corrections to produce incident charged particle ACE files:
- increasing an array size (for the number of Coulomb angles)
- correcting issue #138 in all instances (ACER and GROUPR)

I'm still working on getting a few test files added for the different cases one may encounter:
- MF6 MT2 LAW=5 LTP=1 (nuclear amplitude expansion)
- MF6 MT2 LAW=5 LTP=2 (residual cross section expansion as Legendre coefficients)
- MF6 MT2 LAW=5 LTP=12 (nuclear plus interference distribution with probabilities linear in mu)
- MF6 MT2 LAW=5 LTP=14/15 (tabulation in loglin or linlin)

Still to be added as well: release notes

### DIFF
--- a/src/vers.f90
+++ b/src/vers.f90
@@ -4,6 +4,6 @@ module version
    implicit none
    private
    character(8),public::vers='2016.52'
-   character(8),public::vday='xxNov19'
+   character(8),public::vday='02Nov19'
 end module version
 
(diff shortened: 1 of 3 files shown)


whaeck:
5 tests were added. As expected, only two of them changed from NJOY 2016.51 to 2016.52: the one for identical particles with spin = 0 and the one with spin = 1. The other tests (2 for different particles and one for identical particles with a half integer spin) did not change, as expected.

Release notes and test descriptions were added.

This one is good to go.

src/vers.f90
```diff
@@ -4,6 +4,6 @@ module version
    implicit none
    private
    character(8),public::vers='2016.52'
-   character(8),public::vday='xxNov19'
+   character(8),public::vday='02Nov19'
```
jlconlin: Did you intend for this to be tomorrow?

jlconlin:
By the way, all the tests passed without the updated fortran files.

src/vers.f90
whaeck: Yes, 2016.51 is officially released today.

PR #141, opened 2019-11-07
Added nza again

whaeck:


src/acer.f90
```diff
@@ -358,14 +359,22 @@ subroutine acer
    else if (iopt.eq.2) then
       tempd=300
       tscr=' '
-      nza=16
-      read(nsysi,*) matd,tempd,tscr
+      nza=3
```
jlconlin: I still don't like the default to be 3. While there wasn't an `nza` parameter previously, it was effectively defaulted to 1 as only 1 was required.

whaeck: This is to keep backwards compatible behaviour. If a user has an input like this:
```
    4025 296 'zrzrh' /
    40090 0 0 /
```
and we use nza=1 by default, then the last two zeros may be read as the next input (the mt numbers, number of bins, etc.). Setting nza by default equal to 3 solves this potential issue for old input.

jlconlin: To provide a counter argument what if the old input is this:
```
4025 296 'lwtr' /
1001 /
```
I think providing a single ZA is more common (which is why it was the default).

whaeck: Why do we argue about this? It's a detail.

Anyway, in the case you bring up, the code will read it properly and reset nza to 1. Test 25 covers this behaviour and it passes with the current implementation.

jlconlin: Ah, I didn't realize that it would read correctly (i.e., adding two extra zeros.) If that works, then we are good.

whaeck: I have to mimic the old implementation, in which these values were stored in three different variables and in which the second and third one were defaulted to zero (and that allows a user to add two zeros on the line). It's annoying but that's what it was before.

PR #145, opened 2020-01-21
Added an error message to detect zero awp values for non photons

whaeck:
This pull request "fixes" issue #144 in that it allows NJOY to gracefully exit without a stack trace and provide some information on where the issue is.

It is impossible to recover from this error, unless we assign atomic mass values ourselves. I consider this an error in the evaluation and it should be fixed accordingly.


whaeck:
It would actually be nice if we could also test for failures in our test suite :-)

Issue #130, opened 2019-09-26
Consistency checks introduces spikes in the edges of the secondary energy distributions


jchsublet:
When processing the Fe56 from TENDL 2019, we found a potential problem in the way ACER corrects some data issues in the file while running the consistency checks.

The first issue is related to the epmax value determined by NJOY. Whenever the upper energy of the secondary energy distribution is larger than this value, consis will reset the energy values that are larger to be below this value and "renormalise" the pdf function associated to it. The normalisation does not appear to be correct as this leads to the introduction of artificial spikes in the energy distributions.

Whenever we change to secondary energy distributions to have an upper energy limit equal to the NJOY calculated epmax value, the distributions look as they should, as intended by the evaluator.

A second issue is also related to consis but cannot be corrected in the actual evaluation. ENDF rules (apparently) require MT5 crossections and other related data to begin at 1e-5 eV because the reaction Q value is zero and it is therefore not a threshold reaction. consis also changes the epmax (MF6 MT5 neutron production) given in the COM frame as it should to a LAB frame, but because the data is below the MT5 cutoff, the evaluator only put a dummy distribution here. In this case, it consists of a distribution at an incident energy of 1e-5 eV with secondary energies between 0 and 1e-5 eV. consis wants to change this last value in 9.648869e-6 eV. Suffice to say, this makes little sense.

This is the original TENDL 2019 file: 
[n-Fe056.tendl-19.txt](https://github.com/njoy/NJOY2016/files/3656623/n-Fe056.tendl-19.txt)
This is the corrected file using the epmax value from NJOY: 
[n-Fe056.tendl-19-c5.txt](https://github.com/njoy/NJOY2016/files/3656627/n-Fe056.tendl-19-c5.txt)

This are the input files we used to illustrate the problem:
[inFe056.txt](https://github.com/njoy/NJOY2016/files/3656613/inFe056.txt)
[inFe056c5.txt](https://github.com/njoy/NJOY2016/files/3656612/inFe056c5.txt)

These are the pdf graphs produced by acer for the original and corrected evaluations (after consis made its corrections), see page 57:
[Fe056chk.pdf](https://github.com/njoy/NJOY2016/files/3656631/Fe056chk.pdf)
[Fe056c5chk.pdf](https://github.com/njoy/NJOY2016/files/3656632/Fe056c5chk.pdf)

jlconlin:
[...]

PR #147, opened 2020-01-22
Fix/acer consis

whaeck:
This addresses issue #130 

This relates to changes made by the consistency checks when a checker acer run is requested. NJOY only rarely modifies data. It does so for secondary particle distributions that use LAW=4 (isotropic angular distribution and continuous tabulated energy distributions) or LAW=44 (Kalbach-Mann).

When the consis routine determines the maximum outgoing energy is too high compared to epmax, it sets the last value of the outgoing energy array to this value, and all values before it to monotonically increasing values that are lower than epmax. It then adjusts the pdf values to maintain the associated cdf. This often leads to the introduction of artificial spikes in the plots produced by acer.

This behaviour has now been modified. Whenever this problem is detected, the consis routine will now reset the first energy larger to epmax to a value slightly below it and adjust the pdf value so that the cdf value for this energy point is 1.0. All other energy points after this energy point along with their associated data is set to zero. The locators in the xss array are left unchanged.

Due to this new behaviour, and because the routine that writes the xss array does not check locators, the change routine was modified as well to verify consistency of the locators with the actual position in the xss array. Whenever the locator is inconsistent, it will either just advance to the locator (thus taking into account the gaps created by the consis routine) or issue an error when the locator points to a position lower than the current one.

The change routine was also simplified to improve its readability by adding subroutines to write a list of integers, a list of real numbers, a single integer and a single real number.

A large number of compiler warnings were also fixed in this pull request (mostly unused or uninitialised variables).

Test case 55 was added to properly test this for a LAW=44. Up to now, no case for LAW=4 has been found.

### DIFF
--- a/src/acefc.f90
+++ b/src/acefc.f90
@@ -53,7 +53,6 @@ module acefc
    ! File 6 parameters
    integer::ipnu,jpnu,nkk
    real(kr),allocatable,dimension(:)::e1456,p1456nu
-   !real(kr),allocatable,dimension(:)::p16456nu,p6456nu
    integer::nsix,n16
 
    ! particle production information (from common/ace8/)
@@ -817,8 +816,6 @@ contains
    integer,parameter::maxscr=2000
    real(kr),parameter::xmax=2.e7_kr
    real(kr),parameter::errlim=1.e-5_kr
-   !real(kr),parameter::stpmax=1.25e0_kr
-   !integer,parameter::nc=4
 
    !--initialize
    if (allocated(scr)) deallocate(scr)
@@ -1940,7 +1937,6 @@ contains
    character(60)::string
    real(kr)::dzap,test,zap,e1,e2,f,ei,ep,epn,ss,ff,dmu
    real(kr)::b(50)
-   !real(kr)::y,enext
    real(kr),dimension(:),allocatable::tab1
    real(kr),dimension(:),allocatable::scr
    integer,parameter::nwmaxn=65000
@@ -3869,7 +3865,6 @@ contains
    real(kr),dimension(:),allocatable::scr
    character(4)::blank='    '
    real(kr),parameter::etop=1.e10_kr
-   !real(kr),parameter::emax=2.e7_kr
    real(kr),parameter::zero=0
 
    !--set up size and storage, and assign i/o units
@@ -4526,7 +4521,6 @@ contains
    real(kr),dimension(:),allocatable::egn,egg
    real(kr),dimension(:),allocatable::ee,eb
    real(kr),dimension(:),allocatable::sig
-   !integer,parameter::maxsig=10000
    real(kr),parameter::elo=1.e-5_kr
    real(kr),parameter::ehi=2.e7_kr
    real(kr),parameter::zero=0
@@ -7002,7 +6996,6 @@ contains
    integer::ki,iso,ik3,ii1,ia,ll,intmu,nmu,imu,mus,npep,intep
    integer::last,nx,ix
    integer::jp,jpn,jpp
-   !integer::nxyz1,nxyz2,nxyz3,nxyz4
    real(kr)::test,eemx,yield,xnext,xx,yy,y,xn,eyl,gyl,en
    real(kr)::apsx,step1,step2,xl,pl,yn,pn,rn,sum,ee
    real(kr)::ep,e,bzro,sfe,sfo,bbi,fbarcm,delfcm,akal,rkal
@@ -7015,8 +7008,6 @@ contains
    integer,parameter::nwscr=18000
    real(kr),parameter::emev=1.e6_kr
    real(kr),parameter::small=1.e-30_kr
-   !real(kr),parameter::eps=.001e0_kr
-   !real(kr),parameter::tmin=1.e-6_kr
    real(kr),parameter::etop=1.e10_kr
    real(kr),parameter::up=1.00001e0_kr
    real(kr),parameter::elow=1.e-5_kr
@@ -17248,11 +17239,7 @@ contains
    integer::l,j,kf,n,i,ne,major,minor,it,nr1
    character(1)::qu=''''
    real(kr),parameter::big=1.0e10_kr
-   !real(kr),parameter::small=1.e-12_kr
-   !real(kr),parameter::ten=10.e0_kr
    real(kr),parameter::step=1.2e0_kr
-   !real(kr),parameter::zero=0
-   !real(kr),parameter::one=1.e0_kr
 
    !--set up the page for the total nubar curve
    xmin=big
@@ -17825,8 +17812,6 @@ contains
    real(kr)::xmin,xmax,ymin,ymax,xstep,ystep,x,y,decay,frac,xtag,ytag
    character(1)::qu=''''
    real(kr),parameter::big=1.0e10_kr
-   !real(kr),parameter::small=1.e-12_kr
-   !real(kr),parameter::ten=10.e0_kr
    real(kr),parameter::step=1.2e0_kr
    real(kr),parameter::scale=1.e2_kr
    real(kr),parameter::zero=0


whaeck:
@jlconlin If you don't have enough pull requests, here's another :-)

src/acefc.f90
```diff
@@ -4794,7 +4850,7 @@ subroutine acelod(nin,nedis,suff,matd,tempd,newfor,mcnpx,ismooth)
    use util ! repoz,dater,error,skiprz,sigfig
    use endf ! provides endf routines and variables
    ! externals
-   integer::nin,nedis,matd,newfor,mcnpx,ismooth
+   integer::nin,matd,newfor,mcnpx,ismooth
```
jlconlin: Do we need to include `suff` in this list? (I'm not a Fortran guy)

src/acefc.f90
```diff
@@ -12841,1193 +13011,806 @@ subroutine change(nout)
...
-               call typen(l,nout,1)
-               l=l+1
-               call typen(l,nout,2)
-               l=l+1
+               call write_integer(nout,l)            ! LP
+               call write_real(nout,l)               ! EG
 
             !--law 4 and law 44
+            ! this piece is NOT compatible with LAW=44 (need two additional
+            ! arrays for r and a), potential issue?
```
jlconlin: I don't know if we have test that exercises LAW=44. Why is this a problem?

src/acefc.f90
nathangibson14: `suff` is a `real`, declared on the line below.

jlconlin: So you are right `<feelingSheepish>` I don't know why I missed that earlier.

src/acefc.f90
```diff
@@ -14608,25 +14394,47 @@ subroutine consis
...
+                     if (nnew.ne.nn) then              ! move values if required
+                       do i=1,2
+                         do j=i*nnew+1,3*nn
+                           xss(loci+j)=xss(loci+j+ishift)
+                         enddo
+                       enddo
+                       do j=3*nnew+1,3*nn
+                         xss(loci+j)=0.0
+                       enddo
+                     endif
```
nathangibson14: Didn't work through the math here, but it certainly reads like the description of the changed algorithm you provided in the PR comment.  And it looks about as clean as could be expected, given you had to fit into what was in this file already.

src/acefc.f90
```diff
@@ -12841,1193 +13011,806 @@ subroutine change(nout)
...
+            call advance_to_locator(nout,l,yh)
             nyh=nint(xss(l))
-            call typen(l,nout,1)
-            l=l+1
-            do k=1,nyh
-               call typen(l,nout,1)
-               l=l+1
-            enddo
+            call write_integer(nout,l)            ! NYP
+            call write_integer_list(nout,l,nyh)   ! MTY (NYP values)
```
nathangibson14: It was hard not to fall asleep looking at these repetitive changes.  Props to you for working through all of them (potentially without falling asleep yourself)!

src/acefc.f90
```diff
@@ -14719,25 +14522,50 @@ subroutine consis
...
+                     do j=nnew+1,nn                    ! expunge points above
+                       xss(loci+j)=0.0                 ! Eprime
+                       xss(loci+j+nn)=0.0              ! pdf
+                       xss(loci+j+2*nn)=0.0            ! cdf
+                       xss(loci+j+3*nn)=0.0            ! precompound fraction
+                       xss(loci+j+4*nn)=0.0            ! slope
                      enddo
+                     if (nnew.ne.nn) then              ! move values if required
+                       do i=1,4
+                         do j=i*nnew+1,5*nn
```
nathangibson14: Doesn't this upper limit only need to go to `i*nn`?  Aren't you copying junk from above the limit of this dataset here, then zeroing it out later?

src/acefc.f90
```diff
@@ -12841,1193 +13011,806 @@ subroutine change(nout)
...
+      rlocator=l
+      call write_integer_list(nout,l,ntr)
 
       !--sig block
-      l=sig
+      call advance_to_locator(nout,l,sig)
       do i=1,ntr
-         call typen(l,nout,1)
-         l=l+1
+         call advance_to_locator(nout,l,sig+nint(xss(rlocator))-1) ! sig=jxs(7)
```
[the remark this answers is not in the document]
whaeck: With the change in consis, it is possible that the xss array has a "gap" in them when we expunge them. To solve this, I decided to verify the locators in the xss arrays since the original code here assumed that the entire xss array was continuous (i.e. no gaps). By not doing this check, the start of a gap in the xss array would have been interpreted as the beginning of a new set of data. Since only doing this where I introduced gaps was a bad idea, I decided to perform thes checks for the entire xss array.

src/acefc.f90
whaeck: You have no idea how many times I verified myself to be sure not to screw this up.

nathangibson14: I'm glad, because I have no faith in my review of these, haha.

src/acefc.f90
```diff
@@ -7230,18 +7291,18 @@ subroutine acelf6(next,i,matd,mt,q,iza,izai,nin,newfor,ismooth)
...
-      xss(next+n)=xl
-      xss(next+nn+n)=pl
-      xss(next+2*nn+n)=yn
-      xx=elow
-      do while (xx.lt.test1)
+      xss(next+n)=xl       ! values between zero and one, Eout/E^i_max
+      xss(next+nn+n)=pl    ! probability
+      xss(next+2*nn+n)=yn  ! cumulative probability
+      xx=elow              ! 1e-5
+      do while (xx.lt.test1)  ! as long as xx is smaller than 0.99999
```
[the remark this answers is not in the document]
whaeck: True, these are comments I added trying to understand the code.

src/acefc.f90
whaeck: According to the ACE specifications, there is supposed to be a difference.

src/acefc.f90
whaeck: Yes. But it feels safer to it this way. The xss array we're changing essentially looks like this:
```
x1 x2 x3 y1 y2 y3 z1 z2 z3 f1 f2 f3 g1 g2 g3
```
And we want to make it look like this:
```
x1 x2y1 y2 z1 z2 f1 f2 g1 g2 0 0 0 0 0
```
So, the first time I go through this, the xss array will be:
```
x1 x2 y1 y2 y3 z1 z2 z3 f1 f2 f3 g1 g2 g3 g3
```
The second time:
```
x1 x2 y1 y2 z1 z2 z3 f1 f2 f3 g1 g2 g3 g3 g3
```
and so on.

src/acefc.f90
whaeck: I actually found an error in one of the tests because of this (I had lots of print outs to the screen). One test runs multiple ace files and there appeared to be junk in it because the array was not erased at the end of each run. The locator check was going ballistic over it.

whaeck:
@nathangibson14 The reason why we go from the old treatment in consis to the new treatment was to clean up some of the graphs we were producing. If you had 2 eprime points that were above the calculated epmax (let's say this is 1e-10 MeV), njoy set the last two eprimes to 9.999999999998e-11 and 9.999999999999e-11 (give or take a few 9s). The delta E for these bins is so small that the probability explodes. As a result, the graph only shows spikes. See #130 for an example.

src/acefc.f90
nathangibson14: I'm not worried about the number of flops, although my suggestion would save a few.  But what I am concerned about is what happens when `j=5*nn`.  Your first pass would actually result in:
` x1 x2 y1 y2 y3 z1 z2 z3 f1 f2 f3 g1 g2 g3 JUNK `
Subsequent passes would increase the number of `JUNK` values.

But fortunately, `xss` is bigger than this dataset, so chances of segfaulting are minimal.

nathangibson14:
Aha, I see what I was missing.  The `sigfig` routine is changing the value from `epmax`.  And to be clear, this shouldn't meaningfully change the values produced in downstream calculations sampling, it just makes prettier plots.  I'm happy with that.

whaeck:
One could say it is a cosmetic change with a little more changes behind it ;-)

It did allow me to write up the remainder of the continuous energy ACE specification, maybe I should do that pull request next.

jlconlin:
Do we know if @jchsublet has had a chance to try this and whether or not it fixes issue #130

jchsublet:
@whaeck @jlconlin yes you should know that jchsublet tested parts of this branch as he unadvertantly use it as is njoy2016.53 code when released see my last comment on #130. Having said that I'll be happy to test any other release when it occurs or even a test one

p.s. your timeline is too long between bug report and correction to be efficient

jlconlin:
Thanks @jchsublet We need to get better at speeding up.

PR #157, opened 2020-04-20
Raised THERMR Emax upper Limit from 10eV to 100eV

yunhuangzhang:
We found that the internal limit of Emax<=10eV prevented us from obtaining accurate neutron slowing-down calculations (for instance, in Borated HDPE) in the thermal range. Free-gas and S(\alpha,\beta) treatment were both affected. The initial findings with numerical experiments were documented in our M&C2017 conference paper "NJOY Maximum Energy Limit For Thermal Neutron". A journal article with more in-depth proof based on two-body kinematic analyses is currently under review (2rd round) with Progress in Nuclear Energy.

Raising Emax was achieved by adding energy grid points, which were initially proposed by Dr. A. C. (Skip) Kahler. Raising Emax to 100eV resulted in satisfactory numerical simulations compared to experiments, and we think it should be adequate for most other applications as well. This limit can be further lifted by extending the grid but should be used with caution.

### DIFF
--- a/src/thermr.f90
+++ b/src/thermr.f90
@@ -1508,7 +1508,7 @@ contains
    real(kr)::b,diff,enow,ep,sabmin,tev,ylast
    real(kr)::u,xl,yl,sum
    real(kr)::tone,elo
-   integer,parameter::ngrid=118
+   integer,parameter::ngrid=134
    integer,parameter::nlmax=65
    integer,parameter::nemax=5000
    integer,parameter::mumax=300
@@ -1540,7 +1540,11 @@ contains
      1.42e0_kr,1.55e0_kr,1.70e0_kr,1.855e0_kr,2.02e0_kr,2.18e0_kr,&
      2.36e0_kr,2.59e0_kr,2.855e0_kr,3.12e0_kr,3.42e0_kr,3.75e0_kr,&
      4.07e0_kr,4.46e0_kr,4.90e0_kr,5.35e0_kr,5.85e0_kr,6.40e0_kr,&
-     7.00e0_kr,7.65e0_kr,8.40e0_kr,9.15e0_kr,9.85e0_kr,10.00e0_kr/)
+     7.00e0_kr,7.65e0_kr,8.40e0_kr,9.15e0_kr,10.00e0_kr,&
+     11.00e0_kr,12.50e0_kr,14.00e0_kr,16.50e0_kr,19.50e0_kr,&
+     23.00e0_kr,27.00e0_kr,31.50e0_kr,36.50e0_kr,42.00e0_kr,&
+     48.00e0_kr,54.50e0_kr,61.50e0_kr,69.00e0_kr,77.00e0_kr,&
+     88.50e0_kr,100.0e0_kr/)
    real(kr),parameter::unity=1.0e0_kr
    real(kr),parameter::sabflg=-225.e0_kr
    real(kr),parameter::eps=1.e-4_kr


jlconlin:
Thank you @yunhuangzhang for your Pull Request. This is a fairly straightforward change to the code, but I'm not sure about it's physical relevance.

I'm no expert on thermal scattering, but it seems that 100 eV is much higher than what is reasonable for the thermal scattering treatment. There are a few NJOY contributors (@marquezj, @vedantkm, and @ameliajo) who—gratefully—have helped with these issues in the past. Hopefully they won't mind chiming in on this.

To help us better judge the validity of the physics, can you please attach your M&C paper? I know the other paper is still under review, but we need to have something to review before we can understand the right thing to do.

jlconlin:
Hah,  I just realized that I already had the M&C paper. I've attached it here as a reference.
[MC2017_NJOY_Full.pdf](https://github.com/njoy/NJOY2016/files/4504053/MC2017_NJOY_Full.pdf)

yunhuangzhang:
Thank you @jlconlin for your quick response. Besides the M&C paper, you can find the [article preprint](https://www.researchgate.net/publication/338410350_Analytic_Evaluation_of_Hydrogen_Neutron_Elastic_Scattering_Moments_with_Free-Gas_Thermal_Treatment_Compared_to_NJOY) on my RG homepage, where we used analytic model to justify the need for an increase emax.

The bottom line is, extending beyond the current energy grid does not change any aspect of the code behavior for emax<=10eV. And there are situations where a simulation can benefit from an increased emax.

jchsublet:
A  word of caution, as I remember when we pushed the limit from 4 to 10 eV, thermal scattering belongs to molecular physics and if I could see advantages to push higher for dedicated material in special circumstance I would advice not to rock the boat for reactor physics H in H20 or O in H2O or even O in Pu02, when the s(a,b) mixes with the nuclear fuel first resonances and the beginning of self-shielding in deterministic simulation. Anyway one always can model more than one same material in different regions in simulation: e.g. H in Poly (could be up 100 eV) but O in Pu02 (better be below 10 eV)  a sage precaution

marquezj:
From what I see, the problem is not THERMR itself, but that multigroup codes failback to target-at-rest kinematics when they do not have thermal data. MCNP on the other hand has S(alpha,beta) + free gas at Teff (provided by THERMR + ACER) below a cutoff energy, and free gas at T above the cutoff energy.

Extending the incident energy grid in THERMR should not be a problem, but if it is used for materials other than hydrogen care must be taken not to affect resonances.

@yunhuangzhang did you test this with materials with a coherent-elastic component (graphite, beryllium)? GROUPR has a hard coded limit of 10 eV for Bragg-edge treatment and might complicate things.

yunhuangzhang:
I totally agree with jchsublet and marquezj. Current the two-body model presented in our paper is only based on free-gas hydrogen. Therefore one should be very cautious when applying it to other more sophisticated materials. And yes, a much higher thermal cut-off is not advisable for heavier nuclides due to interference with resonances. As a matter of fact, for those isotopes free-gas treatment has very limited impact on their thermal scattering, and the original 10eV is more than adequate for most cases. 

To be more specific: the proposed fix was verified with H free-gas using the back-to-basic two-body model, and similar trend was also observed from numerical experiment with H poly. I would not advise people to use 100eV as default when dealing with other materials. But it is nice to have the flexibility if one is dealing with H free-gas or H poly.

yunhuangzhang:
> @yunhuangzhang did you test this with materials with a coherent-elastic component (graphite, beryllium)? GROUPR has a hard coded limit of 10 eV for Bragg-edge treatment and might complicate things.

@marquezj No we haven't. We didn't experiment the higher emax with graphite or beyllium, neither did we do any theoretical analyses on them. In our experiments, we only used 100eV emax for the H fg and H poly, and used lower emax for other materials.

yunhuangzhang:
[Analytic evaluation of hydrogen neutron elastic scattering moments with free-gas thermal treatment compared to NJOY .pdf](https://github.com/njoy/NJOY2016/files/4654578/Analytic.evaluation.of.hydrogen.neutron.elastic.scattering.moments.with.free-gas.thermal.treatment.compared.to.NJOY.pdf)

PR #160, opened 2020-04-28
Feature/upn epm

kahlerac:
Adding, and changing some existing, group identifiers in ERRORR and GROUPR.
Share code for uniform group labels in both GROUPR and ERRORR.

@jchsublet Will this affect your legacy NJOY2012 inputs?  We're changing group numbers from the local upn created for the Agency several years ago.

### DIFF
--- a/src/errorr.f90
+++ b/src/errorr.f90
@@ -155,11 +155,11 @@ contains
    !
    ! 9/17/2012 ... important changes
    ! - ngout = input gendf input tape
-   !           -  this tape may contain multiple temperatures,
-   !              multiple Legendre moments and multiple sigma-0
-   !              data, but ...
-   !              - only the first (infinitely dilute) sigma-0
-   !                data will be used.
+   !           - this tape may contain multiple temperatures,
+   !             multiple Legendre moments and multiple sigma-0
+   !             data, but ...
+   !           - only the first (infinitely dilute) sigma-0
+   !             data will be used.
    !
    ! - card 3 is now REQUIRED and the specified temperature must
    !   be one of those on the gendf tape.
@@ -194,8 +194,8 @@ contains
    !    iprint  print option (0/1=minimum/maximum) (default=1)
    !    irelco  covariance form (0/1=absolute/relative) (default=1)
    !
-   !  card 3    *** REQIURED for njoy2012_0917 & later ***
-   !    mprint  print option for group averaging (0=min., 1=max.)
+   !  card 3    (*** REQUIRED for njoy2012 and later ***)
+   !    mprint  print option for group averaging (0/1=min (default)/max)
    !    tempin  temperature (default=300)
    !
    !---for endf/b version 4 (iverf=4) only--------------------------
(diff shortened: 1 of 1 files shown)


src/errorr.f90
```diff
@@ -9569,549 +9591,60 @@ subroutine epanel(elo,ehi,ans,nl,nz,iglo,mfcov)
...
+   !    33      UKAEA  142-group structure   
+   !    34      LANL 618-group structure   
    !
-   !--------------------------------------------------------------------
-   use mainio ! provides nsysi,nsyso,nsyse
-   use util ! provides error,sigfig
+   !-------------------------------------------------------------------
+   use mainio ! provides nsyso
+   use util   ! provides error
+   use groupm ! provides gengpn
```
jlconlin, line 9639: I like how we are now using the same thing in both ERRORR and GROUPR. No need to put this in twice!

src/errorr.f90
```diff
@@ -318,8 +320,23 @@ subroutine errorr
    !      15           sand-iia 640-group structure
    !      16           vitamin-e 174-group structure
    !      17           vitamin-j 175-group structure
-   !      18           xmas 172-group structure
-   !      19           read in, supplemented with endf covariance grid
+   !      18           xmas nea-lanl
+   !      19           ecco  33-group structure
```
jlconlin, line 324: This is going to mess up someone's input. If they update and re-run their calculation, they are going to get different answers. Do we want to avoid using `19` to preserve existing capabilities?

src/groupr.f90
```diff
@@ -2471,6 +2498,1564 @@ subroutine gengpn
      1.252320e+07_kr,1.284030e+07_kr,1.349860e+07_kr,1.384030e+07_kr,&
      1.419070e+07_kr,1.454990e+07_kr,1.491820e+07_kr,1.568310e+07_kr,&
      1.648720e+07_kr,1.690460e+07_kr,1.733250e+07_kr,1.964030e+07_kr/)
+   real(kr),parameter::eg24(282)=(/&
```
jlconlin, line 2501: I'm going to assume that these are the same values that were in `error.f90`.

jlconlin:
PS. I've run the tests on my machine and they all pass. Let's wait and see what GitHub Actions reports.

src/errorr.f90
whaeck: Whenever 19 is used, NJOY2016 will write a message out suggesting changing it to -1. In case the input still assumes the old 19 value, it will crash eventually because it'll try to interpret lines in the input file as the user defined structure.

src/groupr.f90
whaeck: Yes, these were the same

jchsublet:
@wim @jlconlin @kahlerac 
Groups identifiers are extremely important in groupr, why they differs in errorr is for me a mystery, errorr should have a subset, and only a subset with the same read in id. GROUPR is thousand times more important than errorr for useful applications, if you insist to change ERRORR logic do in such a way that it correspond to GROUPR, the influence it will have on legacy input will be minimal and to be honest errorr and errors users are not legion.

Issue #163, opened 2020-05-19
the new suffix entered for checking ACER run is ignored for Thermal Neutron Scattering Data (NSUB = 12)


MAS-OUD:
running the following, with the Thermal Neutron Scattering Data sublibrary in tape 21,
```
acer
-21 -28 0 29 30 /
2 0 1 .00 /
'Processed by NJOY' /
125 500.0 'h-h2o' /
1001 /
222 64 0 0 10.0 /
```
writes `h-h2o.00t` in the first line of tape 29. Running
```
acer
0 29 0 80 90 /
7 0 1 .80 /
'Processed by NJOY' /
```
also writes `h-h2o.00t` in the first line of tape 80, but it must be `h-h2o.80t`.

whaeck:
FYI: I'm not ignoring this issue, I'm still working on the other one you submitted.

I may pick this one up anyway since the binary atomic relaxation fix is giving me unexpected problems.

MAS-OUD:
replacing the following lines of code in `aceth` module (starting from line 599, based on `2016.57` version of `NJOY`)
```fortran
!--adjust zaid
if (mcnpx.gt.0) then
    read(hz,'(f10.0,a3)') zaid,ht
else
    read(hz,'(a9,a1)') str,ht
    if (ht(1:1).ne.'t') then
        read(hz,'(f9.0,a1)') zaid,ht
    endif
endif
if (suff.ge.zero.and.ht(1:1).ne.'t') then
    iza=nint(zaid)
    zaid=iza+suff
    if (mcnpx.gt.0) then
        write(hz,'(f10.3,a3)') zaid,ht
    else
        write(hz,'(f9.2,a1)') zaid,ht
    endif
endif
```
with these ones
```fortran
!--adjust zaid
if (mcnpx.gt.0) then
    read(hz,'(f10.0,a3)') zaid,ht
else
    read(hz,'(a9,a1)') str,ht
endif
if (suff.ge.zero) then
    iza=nint(zaid)
    zaid=iza+suff
    if (mcnpx.gt.0) then
        write(hz,'(f10.3,a3)') zaid,ht
    else
        write(hz,'(a6,f3.2,a1)') str(1:6),suff,ht
    endif
endif
```
solved this issue for me.

PR #170, opened 2020-07-27
Fix/acerzaid

whaeck:
Skip made a few changes to solve issue #163


whaeck:
@kahlerac We should add a test for this, a simple single acer call using a previously generated thermal ace file should do just nicely.

whaeck:
@kahlerac All tests pass on my machine ;-)

PR #172, opened 2020-08-05
Fix/cpt

whaeck:
Fixing an array overflow following NaNs in a charged particle ACE file.

I'm waiting on a response from Mark Paris to see if we can add a test case based on the evaluation (if we can, I'll also make an issue out of it).


kahlerac:
This works, so I approve.  But, "ne" is passed to acecpe so why not make xxs and yys allocatable to the needed size and avoid having to test for an array overflow?

whaeck:
@kahlerac Good point, I looked for where ne came from but I seemed to have overlooked it in the function signature. Making the arrays allocatable and set to a size of ne is probably the best things to do.

Issue #178, opened 2020-10-22
nunx is PURR is ignored


whaeck:
Bob MacFarlane sent us a few updates to the PURR module. One of these updates relates to the nunx value in the input file.

nunx allows a user to ask for all unresolved cross sections at each unresolved energy (nunx=0) or just nunx values. It would appear that this option is actually ignored in NJOY2016 (going back through the source code of NJOY2012 and NJOY99, it would appear that this has always been the case) since nunx is never used in any of the energy point loops.

The value of nunx is used to calculate an integer step value so that the points being calculated are spread out over the unresolved resonance range. E.g. if nunx=2, then nstep=nunr/(nunx-1)=nunr, which means only the lower and upper unresolved energy point will be calculated.

whaeck:
Using test 35 as a template and setting nunx=2 in that input file, PURR crashes on the two do loops (do ie=1,nunr) on lines 435 and 485. This is due to the fact that since nunx=2, only 2 tables have been calculated (for the first and the last unresolved energy point) while these output loops still iterate over every unresolved resonance energy and assume that every energy point has a probability table.

One solution is to change the two do loops in do while loops like the one on line 286 and increment the ie index by the previously calculated nstep.

Another solution is the one proposed by Bob, which is to essentially change the meaning of nunx. In this solution, only the first nunx tables are calculated instead of spacing them out as was originally intended.

whaeck:
Since nunx now controls the number of probability tables we calculate, we'll have to make sure that MF2 MT151 and MT153 are also of the right size so that this does not cause any downstream errors. Using the original nunx provided by Bob is incomplete in this respect. The number of unresolved probability tables is still equal to the number of unresolved resonance points, so all but the first nunx ones are all zero.

whaeck:
We now have a working fix that appears to be accepted properly by ACER and VIEWR. See branch fix/purr-nunx

PR #179, opened 2020-10-23
Fix/purr nunx

whaeck:
This resolves issue #178 

Whenever nunx is set to anything else  but 0 or nunr, purr will now properly calculate the probability tables for the given number of tables spread out over the entire unresolved resonance region, selected from the existing unresolved energy points. The existing MT152 (produced by reconr or unresr - whichever was run last) will be erased and replaced by the corresponding data for the selected unresolved energies. MT153 only contains the generated probability tables.

Test 63 was added to verify this. This test is equivalent to test 35 but with nunx set to 2. Between test 35 and test 63, the first table is numerically the same. The second table in test 63, which corresponds to to the last table in test 35 are not numerically the same - most like due to the random number generator. When test 35 calculates the last ptable, it has already generated 19 tables while test 63 has only generated 1 - thus resulting in a different random number sequence and a difference in the table.

If we wish to ensure that the same table is generated in both cases, we'll have to advance the random number generator by as many numbers as would have been used if all tables were calculated. Any thoughts on this?

As a side note: this PR merges into the develop branch as a staging area before merging into master - to avoid having to introduce a version number while we work through different fixes separately.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -2,7 +2,9 @@
 Given here are some release notes for NJOY2016. Each release is made through a formal [Pull Request](https://github.com/njoy/NJOY2016/pulls) made on GitHub. There are links in this document that point to each of those Pull Requests, where you can see in great details the changes that were made. Often the Pull Requests are made in response to an [issue](https://github.com/njoy/NJOY2016/issues). In such cases, links to those issues are also given.
 
 ## [NJOY2016.62](https://github.com/njoy/NJOY2016/pull/xxx)
-This adds a number of changes to NJOY2016 contributed by the IAEA and CIEMAT. In particular, the following changes were made:
+This adds a number of changes to NJOY2016 contributed by the Toshihiko Kawano, Bob McFarlane, IAEA and CIEMAT. In particular, the following changes were made:
+  - fixed an issue in PURR when nunx is not set to the default value (when nunx is set to anything else but 0 or the number of energy points in the unresolved energy region, purr will now properly calculate the probability tables for the given number of tables spread out over the entire unresolved resonance region)
+  - fixed an issue in RECONR where the wrong channel radius was used for the calculation of the phase shift
   - array sizes were increased
   - additional allocated arrays are now deallocated at the end of the subroutine in which they were created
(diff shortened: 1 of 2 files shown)


kahlerac:
Well, since you bring it up ... Yes I think advancing the random number generator is a good idea.  For example, if someone wants to see the impact of changing from N to N+1 ladders the first N ladders have to remain the same.

.github/workflows/ContinuousIntegration.yml
```diff
@@ -2,11 +2,13 @@
 name: Continuous Integration
 
 # This workflow is triggered on pushes and pull requests to the repository.
-on: 
+on:
   push:
     branches: '*'
```
jlconlin, line 7:
```suggestion
    branches: '**'
```
Two asterisks are needed for branch names that have a `/` in them.

whaeck:
@kahlerac That's also a good idea. We'll probably have to start by allocating a number of random numbers for each energy (e.g. 1000 or 10000 per energy), and each ladder in there. If we set that range high enough, we should be generating the same results for each energy, regardless of changes in number of ladders in previous energies. I prefer not to do this in this pull request though. I'll make a note of it for later.

jlconlin (approved):
I don't fully comprehend the changes that are being made here. However, I trust @whaeck as he does good work. @kahlerac If you think this is good, please approve this Pull Request. If unsure how to do it, please give me a call.

kahlerac (approved):

jchsublet:
@whaeck @kahlerac  I am somehow lost in the mist of it all, could you please point me to the place, a place where I can access your mouture/interpretation of Bob's latest work on Purr? is there a branch for it? on the 2016 code

nathangibson14 (approved):
Confirmed first energy point of Test 35 matches first energy point of Test 63.  Last energy value of 35 corresponds to 63, but data differs, presumably because of the random number generator.  I'm a bit surprised to see the magnitude of difference attributed to the RNG, but I'm not really well-suited to adjudicate it.

I definitely would vote in favor of stepping forward the RNG so that these match in the future, for what it's worth.  But then we'll have to validate Test 35's changed values -- no easy task :)

PR #182, opened 2020-11-14
Coulomb phase shift when APE .ne. APT

toshihikokawano:
Line 3224 in samm.f90, phase-shift should be calculated by the effect…ive radius APE

-                  call pghcou(rho,lsp,bound(i,n,ier),hrx,hix,px,dpx,dsx,&
+                  call pghcou(rhof,lsp,bound(i,n,ier),hrx,hix,px,dpx,dsx,&

### DIFF
--- a/src/samm.f90
+++ b/src/samm.f90
@@ -3221,7 +3221,7 @@ contains
                     ishift(ipx,ier),&
                     iffy,eta,sinphx,cosphx,dphix,jdopha)
                   jdopha=1
-                  call pghcou(rho,lsp,bound(i,n,ier),hrx,hix,px,dpx,dsx,&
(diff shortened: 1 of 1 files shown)


whaeck:
@toshihikokawano Interesting, I've stared at this part of the code a few times already in the last couple of years and never noticed. 

Since the phase shift appears to be calculated using the wrong rho due to this mixup, this might explain the small differences with the new resonance reconstructionimplementation if the Ca40 resonance parameters do define two different radii (which I suspect they do). If that is true, that would also mean that your coulomb wave function implementation that I've been using could have been correct all along.

By the way: I like your avatar picture ;-)

whaeck:
Just verified: Ca40 does indeed have 2 radii defined:
  0.4993153 for calculating P and S
  0.3828792 for the phase shift

whaeck (approved):

PR #185, opened 2021-01-12
Updated Python version to 3.5 minimum

whaeck:
isclose is only available starting from Python 3.5. These changes impose the minimum version 3.5 for the python interpreter.

Since Python2 is no longer officially supported, we make the choice of imposing Python 3.5 as the minimum requirement - even though NJOY2016 does not use python during execution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 
 cmake_minimum_required( VERSION 3.2 )
 
-find_package( Python3 3.5 )
+find_package( Python3 3.5 REQUIRED )
 
 set( CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "Supported configuration types" FORCE )
 


tests/CMakeLists.txt
```diff
@@ -1,10 +1,3 @@
-if ( NOT PYTHON_EXECUTABLE )
```
jlconlin, line 1: How come we are not throwing a fatal error on this anymore?

jlconlin:
I'm glad you caught this. It does need to be updated. I just wonder why you removed the code to throw an error when Python3 isn't found.

tests/CMakeLists.txt
whaeck: find_package should throw an error instead since we require a minimum version.

jlconlin (approved):

jlconlin:
Don't forget to update the ReleaseNotes.

whaeck:
@jlconlin I don't see the issue with execute.py, so I think we're good

jlconlin:
I'm good as soon as the ReleaseNotes are updated. We didn't so anything for the physical constants so that should be done.

whaeck:
You did update ReleaseNotes.md, I'm looking at it right now. Or is there something else you want to add?

jlconlin:
> You did update ReleaseNotes.md, I'm looking at it right now. Or is there something else you want to add?

Yeah. Can you update my brain because I forgot I updated the release notes? 🤦🏻

jchsublet:
With this branch fix/python3.5 NJOY2016.61 make test now works properly, seamlessly and successfully. It was never about the compilation that was fully successful and functional but only the test suite sequence. Test on OsX 10.14 Mojave/ gfortran 8.3.0.

PR #187, opened 2021-01-22
Updating tests to remain the same with NJOY21.

jlconlin:
In an effort to keep the tests between NJOY2016 and NJOY21 the same, I've updated a few of the input decks here in NJOY2016.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,6 @@
 
 cmake_minimum_required( VERSION 3.2 )
 
-find_package( Python3 3.5 REQUIRED )
-
 set( CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "Supported configuration types" FORCE )
 
 project( njoy LANGUAGES Fortran )


tests/61/input
```diff
@@ -7,7 +7,7 @@
  acer
   0 61  0 71 72/             card1
   7 1 1 0.90/                card2
-  /                          card3
+  'e80 h-h2o' /              card3
```
whaeck, line 10: iopt=7 allows you to leave card 3 empty. That's legal. Does lipservice complain about this?

tests/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 set( RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/resources" )
 
+find_package( Python3 3.5 REQUIRED )
```
whaeck, line 3: Did you remove this from the main CMakeLists.txt?

Develop should be updated to the latest master too. I don't think this is done yet.

whaeck (changes requested):

tests/CMakeLists.txt
jlconlin: I did not remove this from the top level. I meant to, but forgot.

tests/61/input
jlconlin: lipservice complains if there is nothing there. We could update lipservice to give a default value. That would be more consistent with the other ways we've been doing things.

jlconlin: An empty string works with lipservice, but no string does not.

PR #190, opened 2021-03-02
Fix/iaea nds

whaeck:
A few array size increases, deallocation of allocated arrays and some additional output suggested by Daniel Lopez Aldama, for the IAEA NDS.

This time in the right branch.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,6 +1,12 @@
 # Release Notes&mdash;NJOY2016
 Given here are some release notes for NJOY2016. Each release is made through a formal [Pull Request](https://github.com/njoy/NJOY2016/pulls) made on GitHub. There are links in this document that point to each of those Pull Requests, where you can see in great details the changes that were made. Often the Pull Requests are made in response to an [issue](https://github.com/njoy/NJOY2016/issues). In such cases, links to those issues are also given.
 
+## [NJOY2016.62](https://github.com/njoy/NJOY2016/pull/xxx)
+This adds a number of changes to NJOY2016 contributed by the IAEA and CIEMAT. In particular, the following changes were made:
+  - array sizes were increased
+  - additional allocated arrays are now deallocated at the end of the subroutine in which they were created
+  - added consistency checking for probability tables in ACER for continuous energy libraries
+
 ## [NJOY2016.61](https://github.com/njoy/NJOY2016/pull/140)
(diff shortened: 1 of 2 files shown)


ReleaseNotes.md
```diff
@@ -1,6 +1,12 @@
 # Release Notes&mdash;NJOY2016
 Given here are some release notes for NJOY2016. Each release is made through a formal [Pull Request](https://github.com/njoy/NJOY2016/pulls) made on GitHub. There are links in this document that point to each of those Pull Requests, where you can see in great details the changes that were made. Often the Pull Requests are made in response to an [issue](https://github.com/njoy/NJOY2016/issues). In such cases, links to those issues are also given.
 
+## [NJOY2016.62](https://github.com/njoy/NJOY2016/pull/xxx)
```
jlconlin, line 4: Don't forget to update `xxx` here.

whaeck: I'm going to pull in quite some changes in develop. I intended to update the release note at every time, and then change that xxx when we actually pull into master.

src/acefc.f90
```diff
@@ -1953,7 +1960,7 @@ subroutine topfil(nin,nout,matd,newfor)
    jpp=0
    jpnu=0
    nkk=0
-   nt1w=4200
+   nt1w=20000
```
nathangibson14, line 1963: Is there logic to these increases, or are these just arbitrary?

src/acefc.f90
```diff
@@ -4457,6 +4469,15 @@ subroutine convr(nin,npend,nout,nscr,nedis,nethr,matd)
...
      (disc(iedis),iedis=1,nedis)
+   deallocate(tot)
+   deallocate(yold)
+   deallocate(scr)
+   deallocate(rr)
+   deallocate(aa)
+   deallocate(yy)
+   deallocate(es)
+   deallocate(eg)
+   deallocate(ee)
```
nathangibson14, line 4480: You've added a lot more deallocates than allocates.  Was this just a memory leak previously?

nathangibson14 (commented):
Looks straightforward, and I don't see any actual computation changing, so this shouldn't even change any answers.

src/acefc.f90
whaeck: Daniel tested a number of preliminary evaluations and has had to increase array sizes as a result. Since he's had to do that, I think we  can expect that to happen for future evaluations so I decided to increase the sizes as he suggested.

src/acefc.f90
whaeck: It's not necessarily a problem or a memory leak, but it has been known to lead to problems if you do not deallocate properly. I believe that, if you do not deallocate an array and then try to allocate it again, you get a segfault (it has happened in the past).

These were also proposed by Daniel and it cleans up the memory so I'm including it.

nathangibson14 (approved):

PR #191, opened 2021-03-09
NJOY2016.62

whaeck:
This adds a number of changes to NJOY2016 contributed by Toshihiko Kawano, Bob McFarlane, IAEA and CIEMAT. In particular, the following changes were made:
  - fixed an issue in PURR when nunx is not set to the default value (when nunx is set to anything else but 0 or the number of energy points in the unresolved energy region, purr will now properly calculate the probability tables for the given number of tables spread out over the entire unresolved resonance region)
  - fixed an issue in RECONR where the wrong channel radius was used for the calculation of the phase shift
  - array sizes were increased
  - additional allocated arrays are now deallocated at the end of the subroutine in which they were created
  - added consistency checking for probability tables in ACER for continuous energy libraries


nathangibson14 (approved):
Release notes are updated correctly.  All changes previously reviewed to get into the develop branch.  Test suite passes for me, GHA passes, so everything looks good to go.

Issue #192, opened 2021-03-17
ERRORR doesn't close files when using the 999 option


jlconlin:
There is an undocumented(?) feature of ERRORR where you can add MF=33 data to an existing Tape. This is done by setting the first argument on Card1 to be `999` (see [Test20](https://www.njoy21.io/NJOY2016/testDescription.html#test-problem-20)). In this case, ERRORR reads in data, and writes data to the file, but it doesn't close the files. 

When run all by itself, everything works fine because Fortran will close the file automatically. However, when running mixed with modern components in NJOY21, the files are not completed and written to disk before the next module is run.

whaeck:
Fixed in #196

PR #193, opened 2021-03-17
Fix bug in ERRORR to close opened files

jlconlin:
This bug fix closes files `nitape` and `notape` in ERROR. It addresses #192.


whaeck:
I think we should pull this in develop first, in case we bump into more problems like this while working on the modern RECONR.

jlconlin:
> I think we should pull this in develop first, in case we bump into more problems like this while working on the modern RECONR.

Changed.

whaeck:
Perfect. I'll wait till the tests finish running but I don't see an issue with this.

whaeck (approved):

PR #198, opened 2021-05-13
Feature/pn iaea

whaeck:
The ACER module has been updated to now use non-isotropic distributions in the photonuclear data libraries.

Additional tests were added using the old photonuclear files and a new IAEA file.

In addition: some common functions were moved to the acecm module. The xss array has now been put in acecm as well.  The writing routines for photonuclear ACE files have been updated to use the testing routines already used for the CE ACE files.

Caveat: there are still some temporary print statements in here which we should remove after testing/validation with MCNP

### DIFF
--- a/tests/71/referenceTape33
+++ b/tests/71/referenceTape33
@@ -14317,7 +14317,7 @@
 /
 1  3/
 '<TENDL-2021 deuteron + Pt184>'/
-'<p>hoton emission for (n,x)     '/
+'<p>hoton emission for (z,x)     '/
 -1 2/
    0.000E+00   2.000E+01   1.000E+01/
 '<e>#L.3H.8]g#HXLX> (<m>e<v>)'/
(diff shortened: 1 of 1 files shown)


jchsublet:
I have re-processed all the 219 targets of the IAEA Evaluated Photonuclear Data Library IAEA-PD2019 with the branch feature/pn-iaea (macOS 10.15, gfortran 9.2) . Compared with the earlier version:

 - for 153 targets the acephn law not coded (law 1 lang 1) and reset to isotropic messages warnings have disappeared
 - No NaN or Inf in either the outputs or the Ace files 
 - Acer check/correct successful
 
 From the evaluation forms side, still 
 
 - 7 targets with messages from ptlegs2  -negative probs
 - 176 targets with messages from acephn  -no heating for recoil in MT 5

Unfortunately the rather large amount of temporary print statements may hide other feature or delicacy. I would be happy to re-test without all those print, this should allow a smooth transition to MCNP validation

whaeck:
@jchsublet I removed the temporary print statements some time ago. You may want to update your version with a git pull.

Still, I might have forgotten a few so I'll go through the code once I get back to addressing Daniel's review comments. So much work, so little time.

jchsublet:
Just pull this afternoon, so you most certainly forgot to remove a major one ! no hurry though

nathangibson14 (approved):
Left some very minor comments.

With the receipt of an internal memo detailing the updates work as expected, this is good to go from my perspective.

Issue #162, opened 2020-05-19
MODER doesn't work for Atomic Relaxation Data (NSUB = 6)


MAS-OUD:
It appears that `MODER` doesn't work for Atomic Relaxation Data (NSUB = 6). In the following, tape 20 is the Atomic Relaxation Data from the ENDF/B-VIII.0 library:
```
moder
20 -21 /
```
or
```
moder
1 -21 /
'mat = 100' /
20 100 /
0 /
```
both fail with the error
```
***error in moder***conversion not coded for mf= 28
```

whaeck:
Hmm, interesting. I'll have a look to see what's going on. Thanks for pointing this out.

whaeck:
Little update on this one:

I have a working branch with an updated moder (fix/moder) but I am seeing issues when using a binary photoatomic file together with a binary atomic relaxation file in ACER. I still need to figure out what's up with that. In ASCII mode, all works normally.

jchsublet:
@whaeck a though from the past, there is absolutely no reasons been physical or numerical to parse atomic relaxation or charged particles evaluations through moder as the ASCII forms of those are more than sufficient for useful application. MODER BCD means binary coded data and was never intended to covers all forms of the nuclear data NJOY's module can handle. It is in fact pretty bold, inutile, likely to have unforeseen consequence to do so.

whaeck:
The remaining issue was a missing repoz(in) call, see the fix/moder branch. It now works properly, both binary and ascii input tapes for ACER give the same ACE file.

PR #199, opened 2021-05-13
Fix/moder

whaeck:
Little update to be able to convert MF28 to binary files. This addresses issue #162

the feature/pn-iaea branch needs to be merged before this one.


tests/59/input
```diff
@@ -0,0 +1,14 @@
...
+21 -31
+acer / Prepare ACE files
+-30 -31 0 27 28                                                 / Card 1
+4 0 1 .31                                                       / Card 2
+'U from ENDF/B-VIII.0 photoatomic data with atomic relaxation'  / Card 3
+9200                                                            / Card 11
+moder
+-30 40
+moder
+-31 41
```
nathangibson14, line 13: There's no check that tape40 and tape41 are actually what they should be.  If these lines are going to stay in the test suite, they should be used.

Running this test myself, I see differences in significant digits (e.g., switch from floating point to exponential) in tape20 and tape40.  Tape21 and tape41 (what's of interest here, of course!) look closer, with the changes looking like switching format but retaining significant digits.  I think that's expected NJOY behavior, so this looks fine.

What are your thoughts?

nathangibson14 (commented):
Left a question on the test case.

This is good to go with a relaxed approach to testing.  But we should make sure we're happy with that relaxed approach.

nathangibson14 (approved):
As I did with the other PR I looked at this morning, I'll approve this now.  If you change the test, I give my blessing to you approving it yourself.  And if you don't, well, it's already approved.

tests/59/input
whaeck: That's the issue with some ENDF files. When using MODER, the file gets reformatted into they NJOY writes its own ENDF files. ENDFtk would essentially lead to the same thing: the ENDF file looks different but it contains the same information. The only way to be sure is to use properly formatted files in the first place which should be reproduced exactly.

I'll replace the original tape20 and tape21 with the reformatted tape40 and tape41. The ACE file might change a bit tough - due to roundoff.

whaeck:
I replaced the input tapes with properly formatted ones. As a result, the ACE files for test 48 and the new test 59 no longer agree within a relative error of 1e-9. They do agree with a relative error of 1e-6, which is essentially the precision in the ENDF file. Tapes 40 and 41 are now also compared (since they should be exactly identical to the input tapes, I modified the CMakeLists.txt file to copy the input files twice instead of creating reference tapes in the test case itself).

Issue #205, opened 2021-06-03
ERRORR covout subroutine only outputs the first subsubsection covariance matrix


whaeck:
NJOY2016 ERRORR is now capable of reading MF34 data with multiple subsubsections in them (PR #206), the associated matrices are also calculated and stored in the temporary tape. For more detailed information, see issues #122 and #203.

However, the covout subroutine only outputs the first l,l1 subsubsection. This still needs to be fixed, and might require additional work in COVR/VIEWR as well.

Issue #203, opened 2021-06-01
Problems processing MF34 for ENDF/B-VIII.0 in 235U and 238U


ocabellos:
Dear NJOY Team,

I have processed MF34 for 235U and 238U in ENDF/B-VIII.0 using NJOY2016.63.
Unfortunately, NJOY2016.63 failed in ERRORR module. 

Firstly, NJOY stops saying: "***error in covcal***storage exceeded in loc."

I have increased "locm" value in subroutine "covcal".

Then, NJOY failed again saying:"***error in covcal***mfcov mt found not equal to input mt."
 
Any idea?? Thanks in advance for your time.

My best regards,
Oscar Cabellos (UPM, Spain)

PD: I attach my two inputs. MF34.txt reads tape21 and tape 31 processed with MF33.txt
[MF34.txt](https://github.com/njoy/NJOY2016/files/6579785/MF34.txt)
[MF33.txt](https://github.com/njoy/NJOY2016/files/6579786/MF33.txt)

ocabellos:
This issue is already in #122...
"ERRORR processing issue for MF34 with multiple subsections"

whaeck:
ENDF/B-VIII.0 U235 and U238 MF34 have been on my todo list for a long time. Now seems as good a time as any to actually look into it.

The issue of not finding the proper MT number, might indicate a problem in ERRORR while reading the ENDF file (there might be an unimplemented format choice that causes ERRORR to not properly read the entire MF38 MT2 section).

whaeck:
@ocabellos I've diagnosed the problem and provided a fix in fix/endf80-mf34 but the resulting covariance matrix is incorrect because only the first of the NI subsubsection is being used in the output (see issue #122 for a more detailed explanation). The matrices associated to each of the subsubsections are calculated and stored in a temporary file though, so we'll still need to fix the covcal routine. Unfortunately, I don't have the time to deal with that right now. I'll open an issue to track that problem separately.

I'm closing this issue and the other one now that the fix is up for merging, and with the new issue to track the covout subroutine issue.

Issue #122, opened 2019-02-14
ERRORR processing issue for MF34 with multiple subsections


whaeck:
A few of the ENDF/B-VIII.0 files have covariance data in MF34 that go beyond a single subsection (that is, not only P0 covariances) in the MT2 section. The ERRORR module crashes on these files because the use of a single section has been hardcoded.

A preliminary investigation of the source code has indicated that, when multiple subsections are present, the ERRORR module crashes when reading the covariance data from the ENDF file. The module loses itself in the ENDF file because it is improperly coded. A first try to correct the coding for reading the covariances appears to break the entire covariance processing.

The ERRORR module reads the covariance data for all types (MF32, MF33, MF34, ...) in a single block of code riddled with if statements that branch depending on the covariance data type instead of specific code blocks for each covariance type. Disentangling this coding will most likely be required to offer a permanent fix.

I will add an input file to illustrate this using one of the ENDF/B-VIII.0 files that has multiple MF34 subsections.

As a side note: ERRORR also appears to only handle MF34 MT2 and ignore any other MT sections that may be present. This is again a hard coded feature.

whaeck:
The following input file illustrates he issue using ENDF/B-VIII.0 U235:
[u235-mf34.txt](https://github.com/njoy/NJOY2016/files/6587648/u235-mf34.txt)

whaeck:
First, a bit of an overview of what the covariance subsections look like.

For MF33, a subsection looks like this:
```
[MAT,33,MT/ XMF1, XLFS1, MAT1, MT1, NC, NI]CONT
<sub-subsection for n =1>
<sub-subsection for n =2>
-----------------------------
-----------------------------
<sub-subsection for n =NC>
<sub-subsection for n =1>
<sub-subsection for n =2>
-----------------------------
-----------------------------
<sub-subsection for n = NI>
```

For MF34, a subsection looks like this:
```
[MAT,34,MT/ 0.0, 0.0, MAT1, MT1, NL, NL1]CONT (MAT1=0)
[MAT,34,MT/ 0.0, 0.0, L1, L11, LCT, NI1]CONTa
[MAT,34,MT/ 0.0, 0.0, LS1, LB1, NT1, NE1/ {Data1}] LIST
----------------------------------
----------------------------------
[MAT,34,MT/ 0.0, 0.0, LSNI1, LBSNI1, NTSNI1, NENI1/{DataNI1}] LIST
----------------------------------
----------------------------------
[...]

PR #206, opened 2021-06-03
Fix/endf80 mf34

whaeck:
Fixing issues #122 and #203.

We're merging this into develop instead of master since the fix is "incomplete": NJOY2016 no longer crashes on U235 from ENDF/B-VIII.0 for MF34 covariance data but we still need to fix covout. The remaining issue is detailed in #205.

@kahlerac this is a simple one, so I'm giving it to you for review. We should schedule a phone call to see if you could help out on issue #205 - I'd do it myself but I'm overcommitted as it is.

### DIFF
--- a/src/errorr.f90
+++ b/src/errorr.f90
@@ -2044,7 +2044,7 @@ contains
   320 continue
    if (iok.eq.0) go to 600
    if (mfcov.eq.34) then
-      if (ld.gt.legord.or.ld1.gt.legord)  go to 650
+      if (ld.gt.legord.or.ld1.gt.legord) go to 650
    endif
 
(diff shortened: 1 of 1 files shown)


kahlerac (approved):
Coding revisions look good, but in the new if block where we now correctly force nc=0 we no longer save n1h which is a flag for LCT (cm/lab system for these data).  We haven't properly saved this variable before but it seems like we might need to know this at some time in the future.

whaeck:
@kahlerac Thanks. About the LCT value, I was wondering about that one myself - as to why it was not being used. It might be that LCT was always set to the same value and that Bob did not bother with adding a treatment depending on LCT.

I'll merge this, and we can debate the LCT value for the next merge on this topic.

PR #213, opened 2021-08-02
Feature/claw

nathangibson14:
Added two points (one low, one high) to the CLAW weight function (iwt=9, 10).  Tests 7 and 8 use iwt=9, so no tests added.


whaeck (approved):
I assume there is a test that uses it in the non-regression suite?

PR #221, opened 2021-09-16
Fix/reconr int6

whaeck:
This is a fun one :-)

ENDF has a "special charged particle" interpolation type INT=6. For more information: see the ENDF manual. To interpolation formula for this special charged particle interpolation type, one needs the threshold energy T.

Like any fortran program worthy of the name, NJOY uses global variables and this threshold energy T (referred to as thr6 in the source code) is one of them. When the terp1 subroutine is called that does the actual interpolation, NJOY uses this global variable. So, what could go wrong?

The answer is a lot. To understand what can go wrong, first some explanation on how NJOY does its job in RECONR:
- linearise every cross section in the ENDF file (including resonance backgrounds) in the lunion subroutine
- linearise resonance reactions with the lunion generated energy grid as a first guess, supplement with resonance energies, etc. in the resxs subroutine
- merge the energy grid and resonance reactions in the emerge subroutine

The issue is in this last one. In the lunion subroutine, thr6 gets set to the proper value for every reaction and is used to linearise the cross section using INT=6 so no problem there. thr6 remains set to the last value it was set to once lunion is done. Now, when emerge goes over the final energy grid, it actually interpolates all cross sections (except the resonance reactions) AGAIN using the new energy grid. However, it does not set thr6 for every reaction so the last thr6 value that was set gets used for every terp1 call. As a result, the interpolation law uses the wrong thr6 value, potentially leading to NaN values.


whaeck:
There seems to be an issue in the Debug build for the fix, looking into that now.

whaeck:
OK, problem solved with Debug mode. Should have put the calculation of thr6 slightly further in the logic.

whaeck:
Let's try again. Now we should be good.

src/reconr.f90
```diff
@@ -4711,6 +4711,16 @@ subroutine emerge(nin,ngrid,ngp,nres,nrtot,iold,inew,nscr)
    go to 510
   220 continue
    if (mth.eq.0) go to 210
+   if (iss.eq.nss) then
+      if (mfh.ne.23.and.awin.ne.zero) awrx=c2h/awin
```
nathangibson14, line 4715: Might be worth adding a comment somewhere about how `iss.eq.nss` is needed for MF10.

nathangibson14 (dismissed):
Thanks for figuring this one out.

Any reason not to capture a test case where this was previously broken but now works?

If you opt to make any changes, feel free to approve them yourself without my review, as they wouldn't be to the functioning code.

Issue #223, opened 2021-11-01
Processing issue for d+Pt184 from TENDL


whaeck:
This issue is essentially issue #184. It popped up again due to the MT11 fixed not being released soon enough. A hotfix release (NJOY2016.65) was released to address.

whaeck:
Resolved with #224

PR #224, opened 2021-11-03
Fix/tendl d+pt184

whaeck:
A small update to include MT11 as a deuteron producing reaction, following issue #223 and #184 

Since this is needed for processing TENDL-2021, we're doing a small release (NJOY2016.65).


whaeck:
Good point, but our biggest concern should actually be the reference files we use in the tests. The current size of the tests folder is 539 MB while the resources folder is only 148 MB.

PR #226, opened 2021-11-18
NJOY2016.66

whaeck:
Release for NJOY2016.66

This has been under development for over a year - see release notes for more information.

### DIFF
--- a/src/vers.f90
+++ b/src/vers.f90
@@ -4,5 +4,5 @@ module version
    implicit none
    private
    character(8),public::vers='2016.66'
-   character(8),public::vday='xxNov21'
+   character(8),public::vday='18Nov21'
 end module version
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -27,7 +27,6 @@ This release fixes a processing issue for the d+Pt184 from TENDL-2021 in ACER. S
 
 This release addresses issue [\#184](https://github.com/njoy/NJOY2016/issues/184) and [\#223](https://github.com/njoy/NJOY2016/issues/223).
 
-
 ## [NJOY2016.64](https://github.com/njoy/NJOY2016/pull/202)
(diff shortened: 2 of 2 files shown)


src/vers.f90
```diff
@@ -3,6 +3,6 @@ module version
    ! These values are updated during the NJOY revision-control process.
    implicit none
    private
-   character(8),public::vers='2016.65'
-   character(8),public::vday='01Nov21'
+   character(8),public::vers='2016.66'
+   character(8),public::vday='18Nov21'
```
nathangibson14, line 7: Does this need to be updated due to my slow review?

whaeck: Well, probably ...

whaeck: What date do you want?

src/acepn.f90
```diff
@@ -1080,8 +1161,15 @@ subroutine acephn(nendf,npend,nace,ndir,matd,tempd,iprint,mcnpx,&
             xss(nex+5)=1
             nex=nex+2+2*2
             xss(last+2)=nex-dlwp+1
+            ! amass=awr/awi
```
nathangibson14, line 1164: Do you really want this commented code in the codebase?

nathangibson14: hashtag valueadded

nathangibson14: It's in at least 3 places.

nathangibson14 (approved):
Fortunately, most of this code has been reviewed previously and has ample validation documentation, so my review was mostly a cursory code style review.  I left a few comments, but they were minor at best.

I'll click approve now, but I'll also run a few more comparisons in ERRORR before you merge.

src/acepn.f90
whaeck: I'm impressed you saw that :-)

I'm still leaving these comments because they represent the formulas for an arbitrary incident particle with a given mass awi. By rearranging and substituting awi = 0, you'll obtain the actual code implemented in these different places. This way, I'll always know where the formulas come from and how they were obtained - which has not been obvious in the past.

src/vers.f90
whaeck: Nevermind.

Issue #211, opened 2021-07-06
Compiler segmentation fault when compiling NJOY2016 with gcc-11


roberto160275:
Hi, I have tried to compile NJOY21 from source as explained here (https://docs.njoy21.io/install.html) and compilation failed due to a segmentation error while doing 'make'. There are many warnings during compilation too. Below is the output from 'make'. I'd appreciate some help to work this problem out.

Greetings,

Roberto.

Vajrayana:bin roberto$ make
[  1%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/locale.f90.o
[  2%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/mainio.f90.o
[  4%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/util.f90.o
[  5%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/endf.f90.o
[  7%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/mathm.f90.o
[  8%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/phys.f90.o
[ 10%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/acecm.f90.o
[ 11%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/acedo.f90.o
/Users/roberto/NJOY21/bin/_deps/njoy-src/src/acedo.f90:30:22:

   30 |    real(kr)::xss(nxss)
      |                      1
Warning: Array 'xss' at (1) is larger than limit set by '-fmax-stack-var-size=', moved from stack to static storage. This makes the procedure unsafe when called recursively, or concurrently from multiple threads. Consider using '-frecursive', or increase the '-fmax-stack-var-size=' limit, or change the code to use an ALLOCATABLE array. [-Wsurprising]
[ 12%] Building Fortran object _deps/njoy-build/CMakeFiles/njoy.dir/src/acefc.f90.o
/Users/roberto/NJOY21/bin/_deps/njoy-src/src/acefc.f90:6964:46:

 6959 |             do ki=1,js
      |                      2                        
......
 6964 |                  +renorm*yys(ki)*(xxs(ki)-xxs(ki-1))
      |                                              1
Warning: Array reference at (1) out of bounds (0 < 1) in loop beginning at (2) [-Wdo-subscript]
[...]

PR #231, opened 2022-03-10
Fix/gcc11

whaeck:
Some corrections in errorr to remove compiler warnings. These were made in an effort to diagnose issue #211. Even though we are not able to resolve the issue, these changes are still worthwhile to preserve.

Several unused variables, uninitialised variables and arrray dimension warnings were addressed.

### DIFF
--- a/src/errorr.f90
+++ b/src/errorr.f90
@@ -2523,9 +2523,7 @@ contains
    real(kr),dimension(:),allocatable::b
 
    allocate(b(nnw))
-   do i=1,nnw
-      b(i)=0.
-   enddo
+   b=0.
 
    !--allocate scratch storage and initialize the spectrum
    !--integral array
@@ -3278,12 +3276,8 @@ contains
 
    allocate(enode(nodmax))
    allocate(b(maxb))
-   do i=1,nodmax
-      enode(i)=0.
-   end do
-   do i=1,maxb
-      b(i)=0.
-   end do
+   enode=0.
+   b=0.
    l3=1
 
    !--read in the resonance parameters for computing cross sections
@@ -3754,25 +3748,9 @@ contains
    allocate(sig(maxe,5))
    allocate(gsig(4,ngmax,6))
    allocate(sens(4,6,ngmax))
-   do i=1,maxe
-      do j=1,5
-         sig(i,j)=0.
-      end do
-   end do
-   do i=1,4
-      do j=1,ngmax
-         do k=1,6
-            gsig(i,j,k)=0.
-         end do
-      end do
-   end do
-   do i=1,4
-      do j=1,6
-         do k=1,ngmax
-            sens(i,j,k)=0.
-         end do
-      end do
-   end do
+   sig=0.
+   gsig=0.
+   sens=0.
 
    lneger=.false.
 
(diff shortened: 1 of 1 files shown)


src/errorr.f90
```diff
@@ -2512,14 +2514,18 @@ subroutine spcint(covm,spc,ne1)
...
-   real(kr)::c1,c2,esp,pe,enext,elow,ehigh,aaa
+   real(kr)::esp,pe,enext,elow,ehigh,aaa
    real(kr),dimension(:),allocatable::scr3
-   real(kr)::b(nnw)
+   real(kr),dimension(:),allocatable::b
+
+   allocate(b(nnw))
+   do i=1,nnw
+      b(i)=0.
+   enddo
```
nathangibson14, line 2528: Can't you just do `b=0` in Fortran?

src/errorr.f90
```diff
@@ -3243,12 +3252,12 @@ subroutine rpxsamm(nwscr,a,ier)
    integer::nsrs,nlrs,njsx,nparb,idis,lord
    integer::i,nb,nw
    integer::isrrr
-   real(kr)::el,eh,spin,ap,awri,aw,fact,en,rat,frac
-   real(kr)::e,enext,wt,wtl,elo,ehi,enxt,ee,eel,tmp
+   real(kr)::el,eh,spin,ap,awri,aw,fact,rat,frac
+   real(kr)::e,enext,wt,wtl,elo,ehi,ee,eel,tmp
    integer,parameter::nodmax=30000
-   real(kr)::enode(nodmax)
+   real(kr),dimension(:),allocatable::enode
```
nathangibson14, line 3258: What's the point of the allocatable combined with a size as a parameter?  Is it just to release memory at the end of the routine?

src/errorr.f90
```diff
@@ -3727,6 +3751,29 @@ subroutine rpxlc0(nwscr,a)
...
+         end do
+      end do
+   end do
+   do i=1,4
+      do j=1,6
+         do k=1,ngmax
+            sens(i,j,k)=0.
+         end do
+      end do
+   end do
```
nathangibson14, line 3775: More of the unnecessary loops.  I think `sens=0.` is much cleaner.  Could even do `sens=0. ! (i,j,k)` if you wanted to highlight that it's not a single float.

nathangibson14 (changes requested):
This is fine as is, but I like some simplified syntax, personally.  If you disagree, I can be persuaded to click approve.

src/errorr.f90
whaeck: This caused a compiler warning, probably because the size was too large compared to some built-in maximum size value. Making it allocatable removes the warning.

src/errorr.f90
whaeck: Yes, you can (I found out after these changes and have used it in the other pull requests). I'll change it.

src/errorr.f90
whaeck: This is the type of message you'd get before the change:
Warning: Array 's0' at (1) is larger than limit set by '-fmax-stack-var-size=', moved from stack to static storage. This makes the procedure unsafe when called recursively, or concurrently from multiple threads. Consider using '-frecursive', or increase the '-fmax-stack-var-size=' limit, or change the code to use an ALLOCATABLE array.

nathangibson14 (approved):

Issue #228, opened 2022-02-02
n-JENDL-5 nxcmax in Acer


jchsublet:
For 59 targets of n-JENDL-5 Acer processing is failing with the error message  ***error in topfil***nxc.gt.nxcmax. Where
nxcmax represent the maximum number of reactions. JENDL-5 evaluators deemed necessary to include many (n,p),(n.d),(n,t),(n,He) and (n,a) partials from CCONE alongside exhaustive mf3-mt5/mf6. 

This would require upgrading nxcmax in acefc.90 from 500 to 750

acefc.f90:   integer,parameter::nxcmax=500

Although this will not stop init or sixbar to complain that particle recoil or distribution are not given or missing.

whaeck:
Fixed in #232 

This is now in develop and will become part of NJOY2016.67 at some point.

PR #232, opened 2022-03-10
Fix/nxcmax v2

whaeck:
Fixing issue #228 and removing some more compiler warnings in acer (making the xss array allocatable, etc.).

This builds upon the gcc11 branch so that one needs to be merged first.


nathangibson14 (approved):
I'm sure if you change the last PR, I'll have to reapprove, but this one looks good.

Allocating the xss array -- crazy!

PR #234, opened 2022-03-14
Fix/jendl5 purr

### LINKED ISSUE #233: JENDL5 issues in PURR
We have received a list of nuclides with issues in NJOY2016 when OpenMC tried to process the new JENDL5 library. See https://github.com/openmc-dev/data/pull/57 for more information.

This issue will track the nuclides and particular issues for PURR.

We've been looking at the following nuclides:
- Mn55
- Sr88 (same issue as Mn55)

whaeck:
A few fixes in PURR to process JENDL5 files.

This resolves #233.

### DIFF
--- a/src/errorr.f90
+++ b/src/errorr.f90
@@ -2523,9 +2523,7 @@ contains
    real(kr),dimension(:),allocatable::b
 
    allocate(b(nnw))
-   do i=1,nnw
-      b(i)=0.
-   enddo
+   b=0.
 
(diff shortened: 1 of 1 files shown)


nathangibson14 (approved):
Same as last PR, I'll have to re-approve.

Is there any way to allocate these variables based on contents of the file rather than just increasing array sizes?  Something that'll prevent us from needing to do this again?  Probably not worth the hassle, I admit.

PR #235, opened 2022-03-15
fix jendl5-acer

whaeck:
When processing JENDL5 U235, ACER crashes while trying to read MF6 MT5. This was tracked down to the multiplicity TAB1 for neutrons being too large compared to the buffer being read in tab1io. Adding the standard do while moreio fixes this and ACER will now run to completion on this nuclide.

### DIFF
--- a/src/acefc.f90
+++ b/src/acefc.f90
@@ -6733,6 +6733,13 @@ contains
          call tab2io(nin,0,0,scr,nb,nw)
          m=nint(scr(5))
          n=nint(scr(6))
+         jscr=1
+         do while (nb.ne.0)
+            jscr=jscr+nw
+            if (jscr.gt.nwscr) call error('acelf5',&
+              'scratch storage exceeded reading lf=1',' ')
+            call moreio(nin,0,0,scr(jscr),nb,nw)
+         enddo
          if (next+2*m+1.gt.nxss) call error('acelf5',&
            'insufficient space for energy distributions',' ')
          jnt=nint(scr(8))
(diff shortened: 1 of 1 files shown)


jchsublet:
@whaeck I am not sure I would agree with all you said above, as I do have a verified Ace file for JENDL-5 U235 with NJOY2016.65, but then I will not argue with LANL on what is or is not a standard MCNP input deck ! having said that and regarding this MF6 MT5 and some other from JENDL-5 evaluations I am curious on how Acer handles the multiplicity's law in lines 541413-15?

 8.020200+4 2.002360+2          0          0          2         339228 6  541413
          5          2         33          4                      9228 6  541414
 1.000000-5 0.000000+0 2.000000+7 0.000000+0 2.000000+7 0.000000+09228 6  541415

I am not convinced that Acer (or even Groupr) has the capacity to linearise the 33 4, does it now? can it be checked with AceTk?

whaeck:
@jchsublet I remember that we talked about JENDL5 so I was rather surprised when Paul Romano told me there were issues with some nuclides - including U235. If you had seen the same problems you would have already told me. I have tried NJOY2016.65 and all modes (Release and Debug, or even the default) seem to cause U235 to fail due to that TAB1 record that is not read in its entirety. It might be a compiler thing? I've been using gcc 9.3. I could see if ifort has the same problem.

About that particular multiplicity. I have to check but I assume it just copies it over (since that is what happens in most places) and then lets MCNP handle it. Now, it's important to note that the specs only allow for histogram or lin-lin so it's probably going to end badly.

kahlerac (approved):
Coding changes look good, so I'll approve this, but in subroutine acelf5 ...
 - any reason to not "fix" the tab1io call in the elseif (lf.eq.7.or.lf.eq.9) block?
 - elseif(lf.eq.5) block isn't used (at the moment), but if it ever is we need to correct the n=nint(scr(5)) typo to be n=nint(scr(6)), and maybe add the moreio fix here also?

nathangibson14 (approved):

PR #239, opened 2022-03-22
Fix/ciemat

whaeck:
This resolves issue #188.

Technical background: when processing the MF6 data, ACER will transform MF6 LAW=7 data into LAW=1 LANG=12 data using an internal ENDF file. The coding for the DLWH block did not loop properly over the transformed data, and as a result generated fully isotropic angular distributions for the secondary particles. The coding for the DLW block on the other hand seems to be correct.

For ENDF/B-VIII.0, only MT16 in n+Be9 is impacted (secondary neutrons and alphas use LAW=7).

Test 72 was added (using ENDF/B-VIII.0 Be9) following the fix.


nathangibson14 (approved):
One line change with proper testing to ensure it worked.  Looks good to me.

Issue #241, opened 2022-05-03
Auto processing of MF10 with subactinide fission in GROUPR


whaeck:
Auto processing of MF10 with subactinide fission in GROUPR fails because the za state to be searched for was improperly set.

This input illustrates the problem: [input.txt](https://github.com/njoy/NJOY2016/files/8613779/input.txt)

PR #242, opened 2022-05-03
Fix/subactinide

whaeck:
Fixes issue #241 

Contrary to what I said before, I propose merging it into develop and release the fix as a part of NJOY2016.67 instead of it being a single update. Some of the JENDL5 fixes broke NJOY as well and we did not release those fixes either.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -2,16 +2,19 @@
 Given here are some release notes for NJOY2016. Each release is made through a formal [Pull Request](https://github.com/njoy/NJOY2016/pulls) made on GitHub. There are links in this document that point to each of those Pull Requests, where you can see in great details the changes that were made. Often the Pull Requests are made in response to an [issue](https://github.com/njoy/NJOY2016/issues). In such cases, links to those issues are also given.
 
 ## [NJOY2016.67](https://github.com/njoy/NJOY2016/pull/xxx)
-This update resolves a number of issues encountered when processing the newly released JENDL5 nuclear data library and TENDL-2021 library for evaluations with sub-actinide fission.
+This update resolves a number of issues encountered when processing the newly released JENDL5 nuclear data library and TENDL-2021 library for evaluations with sub-actinide fission, and provides fixes a few minor issues.
+
+Due to an internal compiler error, gcc-11 will not be supported. Users should consider moving to gcc-12 or any version from gcc-7 to gcc-10.
 
 The following minor changes and fixes are included in this version:
   - several corrections to the ERRORR source code to remove compiler warning messages (these were made while looking into issue #211 for gcc-11 compilation)
   - increased the nxcmax array size parameter in ACER to resolve issue #228
   - increased the nthr array size parameter in PURR to resolve an issue in JENDL5 Mn55 (issue #233)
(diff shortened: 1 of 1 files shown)


tests/72/input.txt
```diff
@@ -0,0 +1,26 @@
+ moder
```
nathangibson14, line 1: Why is this file here?  Looks like a duplicate of `input`, maybe you forgot to delete it.

src/groupr.f90
```diff
@@ -4446,7 +4447,7 @@ subroutine gengpn(ign,ngn,egn)
       ngp=ngn+1
       allocate(egn(ngp))
       do ig=1,ngp
-         egn(ig)=eg28(ig)
+         egn(ig)=eg33(ig)
```
nathangibson14, line 4450: These don't look like compiler warnings but real bugs.  Were we really pointing to the wrong group structure for setting these?  I don't see discussion of this in the issue or the PR.  What's the context?

nathangibson14 (changes requested):
Maybe I'm blind, but I don't see the change that fixes this.  I know it's just a one-liner.  Can you point it out to me for my benefit so I don't lose hair searching for it?

tests/72/input.txt
whaeck: This is a duplicate that I had to create to make the issue, I removed it.

src/groupr.f90
whaeck: The compiler warnings actually told me that eg33 was never used - which is not what we wanted. I must have made this mistake a long time ago and nobody told me ...

whaeck:
> Maybe I'm blind, but I don't see the change that fixes this. I know it's just a one-liner. Can you point it out to me for my benefit so I don't lose hair searching for it?

the two lines that say ```izar=-1``` fix the problem.

ReleaseNotes.md
```diff
@@ -2,12 +2,13 @@
...
 
 ## [NJOY2016.67](https://github.com/njoy/NJOY2016/pull/xxx)
-This update resolves a number of issues encountered when processing the newly released JENDL5 nuclear data library.
+This update resolves a number of issues encountered when processing the newly released JENDL5 nuclear data library and TENDL-2021 library for evaluations with sub-actinide fission.
 
 The following minor changes and fixes are included in this version:
   - several corrections to the ERRORR source code to remove compiler warning messages (these were made while looking into issue #211 for gcc-11 compilation)
   - increased the nxcmax array size parameter in ACER to resolve issue #228
   - increased the nthr array size parameter in PURR to resolve an issue in JENDL5 Mn55 (issue #233)
+  - properly set the za in the case of sub-actinide fission (MF10 MT18 data) when using MF10 automated processing in GROUPR (issue #241)
```
nathangibson14, line 11: Fixing the group structures should be in the release notes.

nathangibson14 (approved):
Hopefully you can fix the release notes without me needing to reapprove, but if not I'll reapprove quickly.

ReleaseNotes.md
whaeck: It'll make the review stale if I push but don't worry about that

Issue #237, opened 2022-03-15
JENDL5 processing of U235 with NJOY2016.66 or lower


whaeck:
Processing JENDL5 U235 fails on ACER in the standard processing path for continuous energy ACE files using NJOY2016.66.

An example input file: [U235.inp.txt](https://github.com/njoy/NJOY2016/files/8255107/U235.inp.txt)

The issue was traced back to the MF6 MT5 section in the evaluation. While processing the data, ACER will read the multiplicity of the first section (stored in a TAB1 record) - which is for neutrons. In this particular section, this multiplciity is a relatively long TAB1, longer than NJOY's tab1io subroutine anticipates. This happens quite often and can be resolved by verifying if the entire TAB1 was read and calling moreio as required. In tis particular part of the code, this was not present. Adding the appropriate lines solves the issue and ACER will produce an ACE file.

Issue #209, opened 2021-06-23
Possible wrong value for qdel


tegaines:
In heatr for tabulated mt458, if no table for END is provided the qdel value is increased by c458(5) on line 734, which for ENDF/B-VIII U235, this value is 8074. However if the END table is provided, qdel is increased by scr(7) which happens to be the number of XY pairs in the upcoming table which ends up being the value of 5; this seems either wrong or a strange value to choose?

The same is done for the EGD and EB table. Is this right?

https://github.com/njoy/NJOY2016/blob/6ef2a1d04ba3cad3c114abaaf6069e317364158d/src/heatr.f90#L721

whaeck:
At first glance, you appear to be onto something and you're probably right. I had a quick look at the code and I have to admit I don't understand how I could have made such a mistake. I blame the reviewer who did not see it either :-)

I'll continue to look into this and will keep you informed.

whaeck:
I had forgotten about this but I have now looked into it and can say that the coding might indeed be incorrect.

qdel represents the delayed fission Q at 0 eV (used for the thermal point representation of this component of the Q value). In the loop on line 693-733, we loop over all tabulated components that are present and read in the various TAB1 records into the scr array. We then store the data in the appropriate arrays. For qdel, when the data is tabulated, I would need the first Q value of the table, which is actually stored in scr(6+2*n1h+2) instead of scr(7). scr(7) actually points to the the NBT,INT data of the TAB1.

PR #243, opened 2022-05-06
NJOY2016.67

whaeck:
This update resolves a number of issues encountered when processing the newly released JENDL5 nuclear data library and TENDL-2021 library for evaluations with sub-actinide fission, and provides fixes for a few minor issues.

Due to an internal compiler error, gcc-11 will not be supported. Users should consider moving to gcc-12 or any version from gcc-7 to gcc-10.

The following minor changes and fixes are included in this version:
  - several corrections to the ERRORR source code to remove compiler warning messages (these were made while looking into issue #211 for gcc-11 compilation)
  - increased the nxcmax array size parameter in ACER to resolve issue #228
  - increased the nthr array size parameter in PURR to resolve an issue in JENDL5 Mn55 (issue #233)
  - fixed an issue for processing JENDL5 U235 in ACER (issue #237)
  - correction in HEATR to properly retrieve the fission energy release component's Q value at 0 eV (issue #209 - there currently is no evaluation impacted by this change)
  - corrected ACER following issue #188. Evaluations using LAW=7 in MF6 data that are added to the DLWH block are impacted by this change (i.e. for secondary particle types that are not the incident particle type). For ENDF/B-VIII.0, only the secondary alphas from MT16 in Be9 is impacted by the change.
  - properly set the za in the case of sub-actinide fission (MF10 MT18 data) when using MF10 automated processing in GROUPR (issue #241)
  - when using the ign option in GROUPR for predefined group structures, some of the newer options did not return the correct structure, this has been fixed

A large number of compiler warnings have now been resolved (unused variables, large static arrays converted into allocatable arrays, etc.). There are still a lot of warnings but their number is heavily reduced. For source files that were corrected in this way, the remaining warnings relate to equality comparisons for real values, unused dummy arguments in subroutines and potential 0 indices into arrays (in all cases, if statements prevented this from happening).

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,7 +1,7 @@
 # Release Notes&mdash;NJOY2016
 Given here are some release notes for NJOY2016. Each release is made through a formal [Pull Request](https://github.com/njoy/NJOY2016/pulls) made on GitHub. There are links in this document that point to each of those Pull Requests, where you can see in great details the changes that were made. Often the Pull Requests are made in response to an [issue](https://github.com/njoy/NJOY2016/issues). In such cases, links to those issues are also given.
 
-## [NJOY2016.67](https://github.com/njoy/NJOY2016/pull/xxx)
+## [NJOY2016.67](https://github.com/njoy/NJOY2016/pull/243)
 This update resolves a number of issues encountered when processing the newly released JENDL5 nuclear data library and TENDL-2021 library for evaluations with sub-actinide fission, and provides fixes for a few minor issues.
 
 Due to an internal compiler error, gcc-11 will not be supported. Users should consider moving to gcc-12 or any version from gcc-7 to gcc-10.
--- a/src/vers.f90
+++ b/src/vers.f90
@@ -4,5 +4,5 @@ module version
    implicit none
    private
    character(8),public::vers='2016.67'
-   character(8),public::vday='12May22'
+   character(8),public::vday='26May22'
 end module version


ReleaseNotes.md
```diff
@@ -1,7 +1,26 @@
 # Release Notes&mdash;NJOY2016
 Given here are some release notes for NJOY2016. Each release is made through a formal [Pull Request](https://github.com/njoy/NJOY2016/pulls) made on GitHub. There are links in this document that point to each of those Pull Requests, where you can see in great details the changes that were made. Often the Pull Requests are made in response to an [issue](https://github.com/njoy/NJOY2016/issues). In such cases, links to those issues are also given.
 
-## [NJOY2016.66](https://github.com/njoy/NJOY2016/pull/xxx)
+## [NJOY2016.67](https://github.com/njoy/NJOY2016/pull/xxx)
```
nathangibson14, line 4: xxx -> 243

src/vers.f90
```diff
@@ -3,6 +3,6 @@ module version
    ! These values are updated during the NJOY revision-control process.
    implicit none
    private
-   character(8),public::vers='2016.66'
-   character(8),public::vday='18Nov21'
+   character(8),public::vers='2016.67'
+   character(8),public::vday='12May22'
```
nathangibson14, line 7: May 26

nathangibson14 (changes requested):

nathangibson14 (approved):

Issue #253, opened 2022-07-20
Automatic GROUPR processing for CINDER


jchsublet:
NJOY2016/GROUPR /10 automatic processing from MF=8 dictionary for CINDER encounters some difficulties when it needs to handle MF=10 channel IZAPs when Z is above 99: 1000xZ + A. The below input illustrates the error: 

outFm247: ***error in getsig*** can't find mf,mt,izar,lfs =        10        4      247    0

when in fact the MF10 MT=4 IZAP is given in the evaluation as 100247 ! it needs to read, handled as I6, printed I8 with a LFS if needed

[inFm247.txt](https://github.com/njoy/NJOY2016/files/9149588/inFm247.txt)
[n-Fm257.tendl.txt](https://github.com/njoy/NJOY2016/files/9149591/n-Fm257.tendl.txt)

Having said that more complex mf3*mf6 mt 5 zam targets and products are properly processed but in the printout when Z is also above 99

 for mf3 mf6 mt  5 zam  992542 (n,x) 99254m1 production
 for mf3 mf6 mt  5 zam 1002420 (n,x) ***** production

whaeck:
This should not be that difficult to correct, I think.

However, the evaluation that you have given here is not for Fm247, it is for Fm257 (I got confused when I tried to find MF10 MT4 in that file and it is not defined).

jchsublet:
Of course, sorry here it is [n-Fm247.tendl.txt](https://github.com/njoy/NJOY2016/files/9156647/n-Fm247.tendl.txt)

whaeck:
The issue seem to be related to line 655 in groupr.f90:
```
         izar=(mfd-((mfd/1000000)*1000000))/10
```
For MF10, the mfd valueis set to the IZA value times 10, so for Fm247 this would be 1002470. The above line would truncate that back to 247. Instead of 1e+6, the above equation should use 1e+7. In the code around this line, 1e+7 is used all the time so I assume this is a typo. It worked so far since we rarely work with elements with a Z above 100.

I made a quick fix in fix/groupr-mf10. NJOY now runs to completion on this case.

jchsublet:
I have tested fix/groupr-mf10 and it runs to completion for all 82 neutron induced cases where it did not before, nice job. However to wrap this properly one would also fix the printout of the product, also 1002470 and 100247m1 in the below particular cases. Outgoing also in i6, not i5 as may be in line 1447,1448
 
[...]

PR #255, opened 2022-07-27
Fix/groupr mf10

whaeck:
Fix for #253


kahlerac (approved):
Looks good to me.  Coupled with the previous positive feedback from JCS, I'd say this is complete.

PR #259, opened 2022-09-11
Fix uninitialized array index in CCCR

kernie:
This fixes a NJOY crash when performing a CCCR run with interface file DLAYXS selected.

### DIFF
--- a/src/ccccr.f90
+++ b/src/ccccr.f90
@@ -3049,7 +3049,7 @@ contains
       nwi=2*nisod
       nwds=mult*nwh+nwr+nwi
       write(idlay)(ha(l1h+i),i=1,nwh),(a(l1r+i),i=1,nwr),&
-        (ia(lli+i),i=1,nwi)
+        (ia(l1i+i),i=1,nwi)
 
       !--delayed neutron precursor yield data by isotope
       l3=l1+nwds


whaeck:
That's an interesting typo. I wonder how long that has been in there.

For QA purposes, could you provide an input file (it should just be a simplified one that demonstrates the crash)? We do not really use CCCCR anymore so I'm not familiar with the input deck for that module so it is probably just easier to ask you.

I also changed the branch that this should go into from master to develop. The develop branch is essentially our staging area for NJOY changes. Once we are ready to release a new version (roughly every 3 months now), we pull all changes into master - and update the version number properly. This way, we ensure that the last commit on master is always equal to the latest version (again, this is just a QA and versioning requirement that we have).

Normally, we also update the release notes in the same pull request to keep track of all changes but since this is a very simple change I can update the release notes in the develop branch once we finish this PR.

nathangibson14:
Thanks for submitting this, @kernie.  As I just told @whaeck offline, I do have a small amount of experience with CCCCR and the DLAYXS file, so I can help to look this over.  But I'll still echo what he said: if you have an input file you can share that shows the issue, that would be helpful for us.

kernie:
Hi,

take the attached input and some U235 ENDF6 file and you should get an error (crash) without the patch.

[input.njoy.u235.cccr.txt](https://github.com/njoy/NJOY2016/files/9554548/input.njoy.u235.cccr.txt)

whaeck:
@kernie Thanks. I've confirmed the crash and the fix. We're good to go here.

whaeck (approved):
This still needs an update in the release notes and the addition of a test but I will do that in a separate PR.

PR #261, opened 2022-09-13
Adding a test for the CCCCR issue

whaeck:


kahlerac (dismissed):
Can't claim this is a rigorous review as my experience with CCCCR is non-existent, but the changes seem reasonable given the comments that were also added to the code.

kahlerac (approved):

PR #264, opened 2022-09-22
NJOY2016.68

whaeck:
This update fixes a number of minor issues:
  - there was an indexing error in the calculation of inelastic thermal scattering mubar in ACER for IFENG=2 ACE files. Test 74 was added to track this issue.
  - there was a division by zero that caused a segmentation fault in Debug mode, also for IFENG=2 ACE files.
  - removed an erroneously added FEND record in the PENDF file when no MF12 data gets added to the PENDF file (related to issue #250).
  - correction in ACER to fix issue #250. Previously it could happen that ACER picked the wrong temperature from a multi-temperature PENDF file when the materials in the PENDF only contained linearised MF3 data (no MF10, MF12 or MF23 data was added by RECONR).
  - fixed an indexing problem in CCCCR for delayed neutron data and added test 76.
  - there was a use of wrong memory addresses in CCCR not accounting for GROUPR's energy structure offset.

### DIFF
--- a/docs/testDescription.md
+++ b/docs/testDescription.md
@@ -374,4 +374,10 @@ Tests 74 was added to track an issue found in ACER while calculating inelastic t
 
 [[input](https://raw.githubusercontent.com/njoy/NJOY2016/master/tests/75/input)]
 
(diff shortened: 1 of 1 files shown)


nathangibson14 (approved):
All changes were previously reviewed, CI passes, and version/date and release notes have been updated.  Looks good to me.

PR #267, opened 2022-10-05
Fix/purr bondarenko

whaeck:
An issue with Bondarenko data written to MF2 MT152 by PURR has been reported by JAEA and an investigation has shown that the Bondarenko data written to MF2 MT152 in PURR is the data obtained from the sampled data directly instead of the data obtained from the probability table directly. There actually was commented code in PURR to this effect but it seems to have not been enabled. 

Tests impacted by these results:
- test 10: MF2 MT152 was changed since the dilutions used in the input are quite low (dilutions above 1 show little difference between Bondarenko data obtained from the sampled data or the probability table because the dilution values are larger than the actual total cross section)
- tests 19, 36, 37 and 38: roundoff in the probability tables not related to these changes


whaeck:
@nathangibson14 poke

jchsublet:
I feel that there has been a misunderstanding in interpreting what physics forms is in a pendf tape after PURR, particularly in MF=3/MT=152. As far as I recall, remember and read from LA-UR-12-27079 "The NJOY Nuclear Data Processing System, Version 2012” [https://mcnp.lanl.gov/pdf_files/la-ur-12-27079.pdf](https://eur01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fmcnp.lanl.gov%2Fpdf_files%2Fla-ur-12-27079.pdf&data=05%7C01%7Cj.c.sublet%40iaea.org%7C3e1be875890a48b72f2908dad69d279d%7Ca2f21493a4d14b7fad07819c824f5c4a%7C0%7C0%7C638058267967581613%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=85ZIYKs82hM7hdDUztFxNltKgtWl%2BzxbAELzLeeUGj4%3D&reserved=0) 

MT=152 Bondarenko-style table, direct sampling, same format as UNRESR
MT=153 Probability table Levitt style, special format

From page 229 to 644 (15 pages) of the NJOY manual, one can read
 and I quote

"During the sampling process, PURR also computes Bondarenko-style selfshielded cross sections just like those produced by UNRESR. These values are printed out so that they can be compared with the results from other methods. For recent versions of NJOY, the Bondarenko self-shielded cross sections replace any previous values on the PENDF file from UNRESR. The Bondarenko cross sections can also be computed directly from the probability table using (501) ….

These values are also printed out. They can be compared to the Bondarenko values from direct sampling to help judge whether adequate convergence has been obtained.

***
When comparing Bondarenko results from PURR with those from UNRESR, several factors should be considered. The PURR results may be more reliable at low sigma0 values than
 UNRESR results because of the more complete treatment of resonance overlap effects, but the unrealistic cross sections in the dips between resonances will eventually make even the PURR results suspect at low values. This effect may be especially apparent for the current-weighted total cross section, which is especially sensitive to the low cross sections between resonances.

***
The code now continues by writing the new Bondarenko cross sections to the output PENDF file using MF=2/MT=152 and the new probability table to the file using MF=2/MT=153.

****
The code then computes and displays the Bondarenko table and the the final normalized probability table. As a cross check it also computes the Bondarenko table from the probability table. It should compare well with the Bondarenko table generated by direct sampling if enough ladders have been used.
"

So yes there is a way to also computes Bondarenko table from the probability table (sigb), but it is not, was not intended to overwrite the Bondarenko-style direct sampling (sigf) always stored in MT=152 in UNRESR style after PURR.

Bondarenko, Levitt tables are two distinct methods to self-shield SLBW URR, the second been more potent

nathangibson14:
In my opinion, there's very little reason to support `sigb` over `sigf` or vice versa.  If anything, the most interesting part of these two methods is in how well they compare.  As the manual states, both are provided in output to allow for checking convergence.

I don't particularly understand use cases for MT152 coming from PURR.  But if one were to want to use MT152 from PURR, I'd presume they would also want it to be consistent with MT153.  For that reason, I do think this change is warranted.  This also will resolve the issue identified by JAEA, an added bonus.

@whaeck and I agree that the real issue here is some other approaches used in PURR related to windowing.  URR processing needs some real attention in the future, and hopefully the NJOY21 project can address this someday soon.

nathangibson14 (approved):
Per my comment in the discussion, I think this is the right change, albeit a mostly inconsequential one.

PR #270, opened 2022-11-01
Extend MATXSR to handle dilution dependent spec and prod arrays of GROUPR

kernie:
In [NJOY2016.49](https://github.com/njoy/NJOY2016/pull/119) GROUPR learned to create MF6 MT18 spec and prod arrays to be stored in the output GENDF file based on issue #118 .

This leads to the situation that the GENDF section MF6 MT18 contains for each group NZ dilution dependent spec and prod values, where MATXSR expected only one value. Consequently, the spec and prod arrays in record 10d of the output file were simply wrong.

This fix lets MATXSR use the dilution dependent value according to the sub-material i.e. the dilution/temperature pair.

Probably solves #168 .

PR #274, opened 2022-11-18
Make sigz array in MATXSR allocatable like in GROUPR

kernie:
When having more than 15 sigz values processed in GROUPR, MATXSR would simply crash because of hard coded array sigz of size 15.

This was now adapted to the GROUPR method, where those arrays are allocatable.

This PR includes already #270 .


kernie:
@whaeck I tested on my machine (Linux gfortran-12) without error... Can I get more error output?

whaeck:
When the CI fails it is often difficult to figure out why. I will give this a try on a machine that has gfortran-9 installed to see if I can replicate the issue.

As a side note: since the arrays that you made allocate-able are local in a function you should not have to deallocate them, but you may want to try an input file with two matxsr runs in them to verify it is fine (in the past I have been bitten when trying to reallocate an already allocated array in a previous module call). In this case I think we're fine but it's better to be sure.

kernie:
Good point. I added statements to make sure those arrays aren't allocated.

whaeck:
Interesting: I've cloned your repo, compiled it with gcc 9.3 and am now running the tests. Test 3 fails (still waiting on test 30 to run) but also test 2 is failing (that's a ccccr test).

whaeck:
Even more interesting, for test 3 I get the following error message with gcc-9.3:

At line 1082 of file /users/wim/njoy-testing/NJOY2016/src/matxsr.f90
Fortran runtime error: Attempting to allocate already allocated variable 'sigz'

Error termination. Backtrace:
#0  0x2adb7173db8b in ???
#1  0x2adb7173f493 in ???
#2  0x401b34 in ???
#3  0x4016ec in ???
#4  0x2adb73334554 in __libc_start_main
	at ../csu/libc-start.c:266
#5  0x40171c in ???
#6  0xffffffffffffffff in ???

I'll update to your latest change - which seems to have solved the issue - from what I can see int he current CI state.

kernie:
So, you were right! Test 3 involves 2 materials and it seems the allocated arrays weren't deallocated even when they were out of scope.

whaeck:
I saw something like that happen before but I wasn't thinking it would actually be the reason behind the errors :-)

Test 2 still fails on my side though. I'll have to look into it some more. SInce it is a CCCCR related thing, it is independent of this PR though.

whaeck:
Test 2 failing is unrelated to this PR so I'm not worried about that in this context. I have to check a few more things, but this might be good to go.

PR #279, opened 2023-02-06
Fixed ERRORR LRF7 segfault.

nathangibson14:
This adds a simple check for whether MF32 is present before adding its contribution to computed covariances.  This only affects LRF=7 evaluations, such as Fe-54 in ENDF/B-VIII.0.

### DIFF
--- a/src/errorr.f90
+++ b/src/errorr.f90
@@ -7439,7 +7439,7 @@ contains
   390 continue
 
    ! add contribution from resonance-parameter uncertainty
-   if (mfcov.ne.34.and.mfcov.ne.35.and.mf32.ne.0) then
+   if (mfcov.eq.33.and.mf32.ne.0) then
       call rescon(ix,ixp,csig,cova,izero,ngn,nmt1)
    endif
 


src/errorr.f90
```diff
@@ -7439,7 +7439,7 @@ subroutine covout
   390 continue
 
    ! add contribution from resonance-parameter uncertainty
-   if (mfcov.ne.34.and.mfcov.ne.35) then
```
whaeck, line 7442: Why not just check for mfcov==33 and mf32 present?

i.e. : if (mfcov.eq.33.and.mf32.ne.0) then

because MF33 is the only one to which we need to add MF32 if it is present.

whaeck (changes requested):

src/errorr.f90
nathangibson14: For an LRF=7 evaluation without MF32 but with MF31, this fix does work.  But now I'm a bit worried about a case where LRF=7 is used *with* MF32.  I'm going to try to figure out what happens in that case.

That said, it's a moot point, as I definitely like checking for 33 better than checking for not everything else.

nathangibson14: I'm fairly confident this makes no difference, but I just updated to your recommendation.  Definitely cleaner code this way.

whaeck (approved):

PR #280, opened 2023-02-07
Fix/fendl32

whaeck:
Added do while moreio where appropriate.

Removed a few warning messages by removing unused variables in errorr.

No test is added for the FENDL3.2 issue - it's just a read issue.


nathangibson14 (approved):
With the ncmax update, this indeed fixes the FENDL issues we were seeing.

PR #286, opened 2023-02-21
Fix/thermr iinc2

whaeck:
A minor fix to THERMR: NJOY will now error out properly when no ENDF file is given (nendf=0) for S(a,b) generation (iinc=2).

This was brought to our attention by email (no tracker item is associated to this fix).

Note: merge this only after #284 is merged.


kahlerac:
Looks good to me ...

PR #288, opened 2023-03-02
Fix/heatr nbt

whaeck:
Some evaluations use MF5 energy distribution data for reactions such as MT16, MT15, inelastic levels, etc. in which more than one interpolation ranges are used. Examples from ENDF/B-VIII.0 are Sn119 and Sn122 (there are more than just these, but these are the ones I looked at).

The integration routines did not extract the interpolation data from the right spot in the ENDF file. The tabbar subroutine (which calculates average outgoing neutron energy) and tabsqr subroutine (which calculates photon recoil) were corrected.


JoffreyDorville:
Thanks for integrating this correction!

kahlerac (approved):
Code changes look good to me.

whaeck:
@kahlerac You'll have to reapprove this since I had to fix a conflict on the release notes.

kahlerac (approved):
Re-approved

PR #289, opened 2023-03-02
Fix/kerma mt102

whaeck:
Essentially two changes:
- the energy of a discrete principal photon is flagged as a negative value, we now take the abs value of the energy to ensure we are using the energy correctly
- Updating the calculation of the recoil energy using photon momentum. The photon recoil was calculated per photon instead of per recoil (in order to obtain the correct value, we have to multiply by the number of photons)

A few tests were updated because they had recoil calculated from photons in them (tests 21, 24 and 26). Changes are minor - and only kerma values are impacted.


kahlerac (approved):
Looks good to me.

PR #290, opened 2023-03-13
max number of photons issue in convr

toshihikokawano:
When the number of levels is N, then the max number of cascading photons is LMAX = N(N-1)/2.
This may exceed the hard-coded number of 100.

### DIFF
--- a/src/acefc.f90
+++ b/src/acefc.f90
@@ -3894,7 +3894,7 @@ contains
 !  imax=49
    imax=50
    imaxsq=imax*imax
-   lmax=100
+   lmax=imax*(imax-1)/2
    nned=50
    allocate(disc(nned))
    allocate(ee(imax))


whaeck:
@toshihikokawano I changed the branch to merge into from master to develop. Develop contains the current release candidate for NJOY2016.70 which should be going into master by the end of March.

This is pretty straight forward so I'll merge this as soon as the tests complete. Is there a particular case that causes the error? I may want to just add that as a test case.

I'll update the release notes accordingly in a separate PR.

toshihikokawano:
Recently we are putting all the discrete angular distributions and photons for charged particle channels in MT 600 - 850, and in some cases the number of photons reaches the max of 100. This happens to relatively light to medium mass nuclei. I will send an example by email.

whaeck (approved):
The change removes the error message in the case sent to me.

PR #291, opened 2023-03-15
NaN produced by acefc/viewr

toshihikokawano:
Occasionally, a dummy particle spectrum may be given at only two energy points in MF6, which may happen for high threshold energy reactions. ACEFC properly processes this case for the photon production, and produces a VIEWR input by the subroutine aplopp(). However, for the particle spectrum, xmin and xmax are not determined in the subroutine aplodd().  Because of xmin=xmax=0, VIEWR produces  NaN in the created PostScript. I copied the block near the line 18233 to the line 17730.

### DIFF
--- a/src/acefc.f90
+++ b/src/acefc.f90
@@ -18724,7 +18724,8 @@ contains
    integer::ipt,mtrh,nmtr
    integer hpd,lsigh,sigh,landh,andh,ldlwh,dlwh,yh
    real(kr)::test,e,xs,xstep,ystep,xtag,ytag,thin,xlast
-   real(kr)::ep,pd,ylast,break,cc,pp,rat,stepm,elast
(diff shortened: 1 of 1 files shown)


toshihikokawano:
I checked the failed test cases 13 and 71, and I'm wondering if the original files could have the problem of Xmax. In the test case 13, the alpha emission energy spectrum goes up to 23.6 MeV at the 20-MeV neutron incident. So, the current NJOY, which sets Xmax = 20, seems an issue.

whaeck:
I'm running your changes now to see where the failures are. If - as one could expect - these are in the viewr outputs, we just need to look at the initial ENDF file to see if they have the issue as well. If that is the case, than we should be fine. If not, it might be a side-effect we need to look into.

whaeck (approved):
OK to merge in a temporary branch so I can update the test results. Only plots are impacted.

PR #297, opened 2023-06-07
Feature/mf7 mt451

whaeck:
This adds MF7 MT451 to MODER. It was tested on a file using MF7 MT451 so it works but I did not include a new test in here since we do not use the data in mF7 MT451 yet.

Updated version and date as well for future release.

### DIFF
--- a/src/moder.f90
+++ b/src/moder.f90
@@ -1066,7 +1066,7 @@ contains
    !--general information (mt=451)
    else if (mth.eq.451) then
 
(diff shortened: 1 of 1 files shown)


kahlerac (approved):
Coding matches the new manual specification recently uploaded by Doro at https://git.nndc.bnl.gov/endf/format/endf6man/-/commit/782fb0cee1ca67702cb41dc0e624b426625988cf.

PR #298, opened 2023-06-12
Fix/groupr li

whaeck:
The fix for the latest GROUPR issue. I looked through the tests, the changes appear to be limited to the 6th significant digit so looks like roundoff.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -5,7 +5,7 @@ Given here are some release notes for NJOY2016. Each release is made through a f
 This update adds the new MF7 MT451 (thermal scattering general information) ENDF format to MODER so that this module will be able to interpret the new MF7 section. No other capability in NJOY2016 currently uses the information in this section.
 
 In addition to this MODER update, this update also fixes the following issues:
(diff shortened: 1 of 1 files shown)


kahlerac (changes requested):
The current lct=0 declaration in Subroutine getfle only works by accident.  A more general fix is to eliminate the local lct declaration and subsequent save lct declaration in Subroutine getmf6.  "lct" has already been declared as global and overlaying a local/save declaration seems to cause old, no longer relevant, data to override the current value of this variable.

whaeck:
I expected this as soon as I found out that the lct variable was a global variable - and not a local one. Do you want to make these changes? We can discuss this in more detail during our next meeting.

kahlerac (commented):
Undid the initial patch and implemented a more general correction.  Also tweaked the text describing the fix.

whaeck:
Well, that was easy :-)

I'll have a look at the tests. I might just revert the test results first to see how this change impacts the tests.

whaeck:
Closing this one in favour of #302

PR #305, opened 2023-08-23
Fix/groupr nubar

whaeck:
Increase some array sizes in groupr to accommodate ENDF/B-VIII.1beta2 U233 nubar (groupr complains about running out of memory).

As a side question: how many points does it take to represent nubar?
In the case of ENDF/B-VIII.1beta2 U233, the answer is 8791 points.


jchsublet:
I like the side question, so may be you should ask the evaluators why such structures? as I doubt it could be experimentally based

[tape32.pdf](https://github.com/njoy/NJOY2016/files/12431447/tape32.pdf)

kahlerac (approved):

PR #311, opened 2023-09-26
Fix/mt typo

whaeck:
A typo was found in the name for MT195. It said n,2n2a instead of n,4n2a.


kahlerac (approved):
Checked the manual and agree with this name change.  Checked gaspr.f90 and it seems to have the correct multiplicity data already so no other change needed.

PR #315, opened 2023-10-12
Fix/leapr segfault

whaeck:
This PR fixes a few issues in LEAPR:
- when the phonon expansion order is quite large, a segfault occurred due to the sice of an array. That array size was increased. In addition, additional output is written to the standard output stream to inform the user of calculation progress.
- when the beta grid contains very small values, the trans subroutine calculates a delta value of roughly that size. For a value of for example 1e-13, an inifite loop can occur because the updated beta value (beta+delta) remains the same when beta is large enough. When that happens, we now increase the delta value to compensate.

test 80 was added to verify this last issue.

PR #314 must be merged prior to this one.

### DIFF
--- a/src/leapr.f90
+++ b/src/leapr.f90
@@ -1234,7 +1234,10 @@ contains
       else
          sb(i)=0
       endif
-      if (bet.eq.(bet+delta)) delta=delta*10
+      ! if delta is to small for the current value of beta, increase it
+      do while (bet.eq.(bet+delta))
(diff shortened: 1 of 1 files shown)


kahlerac (approved):
Nothing wrong with a larger fixed dimension for maxt but in the unlikely event another increase is ever needed consider just allocating maxt to a length of nbeta.  Everything else looks fine.

PR #316, opened 2023-10-12
Fix/intel

whaeck:
A few fixes reported by Dan Roubtsov (CNL, Canada).

PR #315 must be merged prior to this one.


kahlerac (dismissed):
thermr change is ok but tricky.  if the commented code following 170 continue is ever restored then rat is no longer unused ... but the compiler will tell you that!

PR #319, opened 2023-10-26
Feature/kbk

whaeck:
This PR adds the different rmatrix background options to the samm.f90 coding. The background data is stored in a new array and the rmatrix is modified when required (subroutine setr). All three options (arbitrary tabulated complex function, Sammy and Frohner parametrisation) are supported.

This change is made in anticipation of a new Sr88 evaluation that will make it into ENDF/B-VIII.1 that uses the background rmatrix option (the Sammy parametrisation in particular). The test results have not been compared to Sammy data (I'm waiting on ORNL feedback) but that does not have to stop us to start the review process.

Little caveat: the use of these background options requires updates to the partial derivatives - which are used by ERRORR when MF32 covariances are used. I updated dead code that seemed to add these derivatives for the Sammy parametrisation but none of the other options have any logic in the code to calculate these derivatives. Since the original code was never used, I'm also not sure of the results. Luckily for us, the Sr88 evaluation currently does not have covariance data associated to them.

I also updated the code to add a number of error messages:
- the LRF=7 code only works with KRM=3, any other value triggers a call to error
- IFG != 0 is now flagged as an error (the coding assumes full widths)
- background rmatrix elements are only allowed for KRM=3, as per ENDF rules

For more information on these options: https://www.nndc.bnl.gov/endf/format/endf-manual-latest.pdf


whaeck:
Comparisons with Sammy have shown that the calculated cross sections are correct within ENDF precision (there is only one channel in the test file: elastic):
Figure 3
Figure 6
Figure 8

whaeck:
There was another correction in the sammy data, so the capture ratio is like this now:
Figure 6(1)

Also seems to be within ENDF precision so we're good on all fronts.

kahlerac (approved):
Impressive work!

PR #320, opened 2023-10-26
Fix/leapr array

whaeck:
@HunterBelanger Everything looks good now. Tests pass so we're good to go. The failing tests must have been fixed in the last update you made.

### DIFF
--- a/src/leapr.f90
+++ b/src/leapr.f90
@@ -423,7 +423,13 @@ contains
    isym=0
    if (ncold.ne.0) isym=1
    if (isabt.eq.1) isym=isym+2
-   mscr=4000
+   
+   ! Based on endout, to write the actual TSL data, the max number of entries
+   ! needed in scr is either 8+2*nalpha, or 8+2*nedge. However, we have no way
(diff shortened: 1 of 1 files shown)


kahlerac (commented):
Surprising to not see any deallocation statements in subroutine contin.  contin is called from within a loop over user specified temperature.  Do existing tests include a multiple temperature leapr job to assure we don't get an already allocated runtime error?

HunterBelanger:
Test 23 should have multiple temperatures in the LEAPR input. I believe that since Fortran 95, allocatable arrays are automatically deallocated when they go out of scope. I think this is why there are no deallocation statements at the end of contin. I didn't see any for the other arrays like `p`, `tlast`, `tnow`, and `xa`, so I didn't add one for `maxt`.

kahlerac (approved):

whaeck:
Deallocation will be required when you run LEAPR multiple times in the same input deck. This is an issue we have seen in the past with other modules like ERRORR. In this case, I don't think it is necessary to protect users against themselves.

As such, I'll pull this into develop. If the deallocation comes up later, we'll just fix it then.

PR #325, opened 2023-12-14
Fix/euler

whaeck:
An error was introduced so many years ago in the value of the Euler-Mascheroni constant when a number of physical constants were put in the phys.f90 file. 

Luckily for us, the impact is limited. The constant is used in the calculation of the Coulomb wave functions - I think - so this will only be seen on nuclides that have charged particle channels in their resonance parameters (LRF=7 only). In the tests only test 20 is impacted as this is the only LRF=7 evaluation we use for our non-regression tests (again, I think). The elastic scattering and capture cross section only have a few values where the 6th digit after the decimal point is different. In MT103 and MT600 (the n,p reaction calculated from the resonance parameters), the impact is only really visible in low cross section values where the 4th to 6th digit after the decimal point are different. This is reflected in the few changes in the total cross section.


kahlerac (approved):
Looks good ... and sadly we had the correct value(s) when this was explicitly typed in multiple code locations!  It was only when moving to phys.f90 that the typo occurred, :(.

whaeck:
I know. I verified it to see when it was introduced.

PR #329, opened 2024-02-07
Feature/sza

whaeck:
This update made a small change to the NXS array for continuous energy and photonuclear ACE files (iopt = 1 and 5). In both cases, the isomeric state S, atom number Z and mass number A are now stored in NXS(9), NXS(10) and NXS(11) respectively.


kahlerac (approved):
Looks good, sorry I missed the initial review request.

PR #330, opened 2024-02-08
Feature/zaoption

whaeck:
This adds an option for the zaid of continuous energy and photonuclear data ACE files:
      - use za in the zaid regardless of the metastable state of the nuclide (default, this is how NJOY has always worked)
      - use the metastable zaid rules (MCNP 6.3 or lower) in which we use za for ground state nuclides and use za + 300 + s * 100 for metastable nuclides (Am242 and Am242m are exceptions to this rule, for these we use 95642 and 95242 respectively)

In the future, we will use this new option to set up the new ZAID for xsdir 2.0 (e.g. Am242m1_293K_ENDF71 instead of 95242.81c).


kahlerac (dismissed):
... and another example of why users should always terminate an input line with a slash!

whaeck:
> ... and another example of why users should always terminate an input line with a slash!

And I keep telling them that ;-)

Issue #76, opened 2018-04-03
thermr typo in effective temperature values


ameliajo:
The gateff function in thermr is used to look up effective temperatures from a GA report ("Reference Manual for ENDF Thermal Neutron Scattering Data" by Koppel and Houston, reachable at https://www.osti.gov/servlets/purl/4075168).

On line 579 of thermr.f90, we have 
`     1064.e0_kr,1220.e0_kr,1229.3e0_kr,&
`
where 1064 is the mat number (corresponding to Be), 1220 is the temperature, and 1229.3 is the effective temperature. It looks like this was a typo, and that the 1220 should really be 1200. 
The Be information is in Section 6 of the aforementioned document. Doesn't seem like this makes too much of a difference but perhaps worth mentioning.

PR #334, opened 2024-04-19
Fix/typo thermr

whaeck:
See #76


kahlerac (dismissed):
Verified per ENDF-356.

PR #343, opened 2024-08-01
Increasing array size

whaeck:
Just a simple fix: increasing the size of internal arrays to ensure that VIEWR runs to completion.

### DIFF
--- a/src/viewr.f90
+++ b/src/viewr.f90
@@ -1323,7 +1323,7 @@ contains
          enddo
          i=i+2+2*nn
          j=j+1
-         if (j.lt.length) then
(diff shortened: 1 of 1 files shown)


kahlerac (approved):
OK, but maybe next time (there's always a next time!) make the 2000 a parameter variable and add a test after line 1324 (j=j+1) to avoid a possible array overflow.

kahlerac (dismissed):

whaeck:
@kahlerac That's indeed better, I made the change you suggested. You'll have to reapprove though.

kahlerac (approved):

PR #346, opened 2024-09-22
(Not necessarily expecting a merge) Unbiased Chi2 sampling in PURR

gridley:
I found in my PhD studies that the Chi^2 sampling scheme in PURR is pretty severely biased. It uses a pre-tabulated set of twenty chi^2 samples for one through four degrees of freedom. Obviously, choosing these at random leads to a mean of the sample mean rather than the true mean of 1, 2, 3, and 4.

This PR is partly just to make this known.

I tried generating a whole nuclear data library using this patch and couldn't pick up on any appreciable differences between the original version of the code and this one. This appears to be the case because despite the samples being biased, the widths sampled according to Chi^2 R.V.'s always appear as ratios, so the ratio terms themselves are not biased. Moreover, using the ptables as a stochastic dilution factor for pointwise grid is also normalizing out potential biases. Perhaps this would explain some of the historical disagreement between NJOY and PREPRO as Red Cullen has pointed out before, suggesting that the samples should be normalized and treated as a multiplying factor. Of course, there are other reasons for this such as decusping the interpolation scheme. I can't find the paper at the moment but will find it if anyone requests.

If you plot the CDF of the cross section distribution, you *can* observe for many nuclides that PURR is missing the tails by quite a bit. The small set of samples essentially fails to hit rare events where the widths are very large or very small, and this governs the tails of the cross-section distribution.

Here is a plot of U235's XS multiplier term at its first URR energy point. You can see the tails are cut off by the default implementation.

image

Here is the U238 total XS multiplier at the first URR energy point with and without this fix. Again it is a small effect but visible in the tails.

image

*maybe* there would be some discernible difference for deep shielding problems of neutrons in the URR range, but aside from that, I'm not sure this makes much of a difference.

I'm just raising this PR for posterity to see, in case someone else ever gets worried about the strange Chi^2 sampling method in PURR.

### DIFF
--- a/src/purr.f90
+++ b/src/purr.f90
@@ -1698,31 +1698,8 @@ contains
    ! internals
    integer::nr0,ir,idone,ndf,i,n
    real(kr)::dcon
-   real(kr),dimension(20,4),parameter::chisq=reshape((/&
-     1.31003e-3_kr,9.19501e-3_kr,.0250905e0_kr,.049254e0_kr,&
-     .0820892e0_kr,.124169e0_kr,.176268e0_kr,.239417e0_kr,&
-     .314977e0_kr,.404749e0_kr,.511145e0_kr,.637461e0_kr,&
-     .788315e0_kr,.970419e0_kr,1.194e0_kr,1.47573e0_kr,&
-     1.84547e0_kr,2.36522e0_kr,3.20371e0_kr,5.58201e0_kr,&
-     .0508548e0_kr,.156167e0_kr,.267335e0_kr,.38505e0_kr,&
-     .510131e0_kr,.643564e0_kr,.786543e0_kr,.940541e0_kr,&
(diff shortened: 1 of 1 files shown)


whaeck:
I wanted to reply to this but other things needed my attention. I looked through this and it looks solid to me. Interestingly enough, this issue was brought to my attention by another person around the same time you put up the PR. There's probably more of these things in the source code. It might have made sense over 30 years ago, but not today.

I'll run a few more tests over the next few weeks and will most likely merge this into a feature branch on this repository, prior to merging it into the develop branch - which is the staging area for new releases.

gridley:
Just pinging you here @whaeck. Any thoughts on the matter since you last looked?

whaeck:
I'm still planning on integrating this into NJOY2016. I have been traveling a lot in November and December so I have had little time to continue working on this.

PR #370, opened 2025-07-30
HEATR issue correction related to the recoil energy calculation

dlopezaldama:
Chen’s recommendation is applied to calculate the recoil energy for charged particle reactions (MT=103, 104, 105, etc.). This means that now the available energy Ea is used in all cases, rather than taking the minimum between Ea and the Coulomb barrier Ec.
Additionally, an error was identified in the treatment of recoil damage for MF=6/LAW=7 (irec>0). 
It does not affect ENDF/B-VIII.1 library, but affects U-235 and U-238 from FENDL-3.2c library
(Here, the data at low energy is taken from ENDF/B-VII.1).


dlopezaldama:
@whaeck, okay, we'll keep in touch.

Issue #168, opened 2020-07-09
Inconsistency of fission matrix in the matxs file


akioyamamoto94:
**Outline**

When some of fissionable nuclides (e.g., Cm246 in JENDL4.0) are processed by NJOY2016 using a standard multi-group generation procedure (i.e., RECONR, BROADR, UNRESR, THERMR, GROUPR, MATXSR), the generated matxs file contains inconsistency in the fission matrix appeared in the block “10d”.

See the attached file for the details.
[20200708_IssueMatxsNjoy.pdf](https://github.com/njoy/NJOY2016/files/4899765/20200708_IssueMatxsNjoy.pdf)

PR #371, opened 2025-07-30
Fixes #369 : MATXS formatted tape correction for constant spectrum and production

### LINKED ISSUE #369: MATXS formatted tape for constant spectrum and production for NZ>1
There is a bug in matxsr.f90 that incorrectly formats the constant spectrum vector (spec) and production vector (prod) when the number of Bondarenko background cross sections (sigz) is greater than one (NZ > 1).
This issue was reported by Ivo Kodeli, who encountered it while processing Pu-239 from JENDL-5. 
However, the bug is systematic and could affect any evaluation where NZ > 1 for the magnitudes spec and prod.

dlopezaldama:
Solve the issue related to the constant spectrum vector and production vector (record 10d on MATXS formatted files) when these magnitudes show a sigz (Bondarenko sigma0) dependency on the GROUPR formatted output tape.


kernie:
See #270 and #274.  At the time I was sure it would solve #168, this has to be checked, if our PRs ever make into NJOY...

PR #391, opened 2026-01-20
ENDF/B-VIII.1 Discrete primary gamma format discrepancy handling in ACER

amlewis1578:
Fix in `ACER` for the ENDF/B-VIII.1 format discrepancy for discrete primary gammas in File 6 - the energy listed is the outgoing gamma energy. Previous JENDL releases listed the binding energy, and that is what NJOY2016 expected. This change gathers the library and release number and for ENDF/B-VIII.1 only, converts the energy to what NJOY2016 expects. It also adds a new test case (87) that uses the ENDF/B-VIII.1 Pt190 evaluation.

The library and release number is checked because it is likely that the next release of ENDF/B will resolve this issue.

The corresponding manual change merge request is [here](https://github.com/njoy/NJOY2016-manual/pull/35)

**Note:**
This commit doesn't touch `HEATR` - that is a separate problem.


src/acefc.f90
```diff
@@ -45,6 +45,11 @@ module acefc
    integer::mt103,mt104,mt105,mt106,mt107
    integer::mpmin,mpmax,mdmin,mdmax,mtmin,mtmax,m3min,m3max,m4min,m4max
 
+   !-- Modification Jan 2026 by A. Lewis
```
whaeck: I think you can remove the modification comments (begin and end)

src/acefc.f90
```diff
@@ -8669,6 +8686,23 @@ subroutine acelpp(next,matd,ngmt,nin)
...
                      endif
+                     !-- Modification Jan 2026 A. Lewis
+                     !-- ENDF-8.1 formats primary gamma energy as the outgoing
+                     !-- energy at each incident energy. JENDL (and JEFF) format
+                     !-- them as the binding energy. NJOY expects the binding 
+                     !-- energy so this logic changes the value if the evaluation
+                     !-- is from ENDF-8.1. This will hopefully be resolved before
+                     !-- the next ENDF release. 
+                     if (law.eq.1.and.ep.lt.zero) then 
+                        if (nlib.eq.0.and.nver.eq.8.and.lrel.eq.1) then
```
whaeck, line 8690: Maybe add a warning for the user to check the format to be sure it is done correctly for non-ENDF evaluations?

whaeck:
I updated the release notes in the develop branch. You should pull that in, and update the release notes for this development.

src/acefc.f90
amlewis1578: they've all been removed

src/acefc.f90
amlewis1578: I added a message in the else block that the gammas are assumed to follow the JENDL format

whaeck (dismissed):

Issue #277, opened 2023-01-06
Problems with JEFF-3.3 Cr50, Cr53, and Cr54


HunterBelanger:
I recently processes all of JEFF-3.3, an upon trying a few benchmarks, I noticed that Cr50, Cr53, and Cr54 all had a similar problem in the generated ACE files. Each of these evaluations has an MT 5, for which the distribution is given in MT 6 as LAW 1 LANG 2 (Kalbach Mann). For each of these evaluations, I find that for at least one incident energy, the out going energy grid is not sorted. This also leads to the CDF for the outgoing energy to have negative values.

As an example, using Cr50, this problem occurs at an incident energy of 4MeV, for which I am seeing an out going energy grid of:

```
Eout = [0.001, 3.71661e-05, 4.4198e-05, 5.25604e-05, 6.25049e-05, 7.4331e-05, 8.83946e-05,
        0.000105119, 0.000125008, 0.00014866, 0.000176787, 0.000210236, 0.000250013,
        0.000297316, 0.000353569, 0.000420465, 0.000500018, 0.000594623, 0.000707127,
        0.000840917, 0.00100002, 0.00118923, 0.00141423, 0.0016818,
        0.002, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1, 1.1, 1.2, 1.3, 1.4,
        1.5, 1.6, 1.7, 1.8, 1.9, 2, 2.2, 2.4, 2.6, 2.8, 3, 3.2, 3.4, 3.6, 3.8, 4]
```
From this grid, we see the that the exit energy grid isn't sorted due to what appears to be the very first energy point. Looking in the evaluation however, I found that the given outgoing energy grid is supposed to be

```
Eout = [0.001, 0.002, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1,
        1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2, 2.2, 2.4, 2.6, 2.8, 3, 3.2, 3.4, 3.6, 3.8, 4]
```

[...]

Issue #131, opened 2019-09-26
ismooth operations on MT5 secondary energy distributions


jchsublet:
MT5 is essentially a change in how reactions are disentangled: starting at an incident energy of 20 or 30 MeV, all reactions are lumped together into a single reaction with multiplicities and secondary distributions.

The smooth operation will add a sqrt(E) tail to the lower energy part of the secondary distribution up to the threshold energy of the reaction. This is justified for the explicit reactions (MT2, MT51, etc). Since MT5 requires to be defined starting at 1e-5 eV, it does not seem necessary to extend the secondary distributions this low for incident particles with energies above 20 or 30 MeV. The evaluator already has to cut off the distribution anyway.

It may be a good idea to exclude MT5 from automatic ismooth operations.

jlconlin:
I guess I'm not sure what you are suggesting here. Is this simply to exclude MT=5 from being smoothed?

whaeck:
@jlconlin Yes, this would just exclude MT5 from ismooth.

I suggest we add an option to allow users to switch it on again.

jchsublet:
This option would need to be set by default to ismoo=0 for mt=5
MT=5 correspond to (z,anything), with each particle identified with multiplicity and spectra given in file 6. As its Q value is set to 0.0, secondary would need to extend to 1e-05 eV, although it has a threshold, 20 or 30 MeV when used as a format transition framework

whaeck:
Addressed in #409

PR #404, opened 2026-06-23
Fix bug in ACER when smoothing distributions

HunterBelanger:
There is a bug in the smoothing operation of ACER which leads to outgoing energy grids from Kalbach-Mann distributions that are not sorted. It has been reported in issues #277 , #131 , and in [LA-UR-18-24034](https://nucleardata.lanl.gov/files/la-ur-18-24034.pdf) when describing the problems with Ne20 for lib80x. The smoothing operation always assumes that the first tabulated outgoing energy is 0 eV, and it then goes about inserting extra points between the first and second provided outgoing energies. However, there is no check in the code that the first energy is 0 eV, and when the first tabulated outgoing energy is not 0 eV, we end up with outgoing energy grids which are not sorted, and cannot be reliably used for sampling. If the evaluator explicitly cut-off the distribution at some given energy above 0 eV, it does not make much sense to go about smoothing the distribution below their lowest provided outgoing energy. As such, this fix adds a check where smoothing is only applied if the first outgoing energy is 0. Otherwise, smoothing is skipped for that particular portion of the distribution.

With this fix, I am able to read the generated ACE files for JEFF-3.3 Cr50, 53, and 54 without a problem. I can also read the files from JEFF-4.0 U237 and B11 without an issue as well. There are many other problematic nucleides which should be fixed by this as well (like Ne20 in ENDF/B-8.0 and 8.1).

### DIFF
--- a/src/acefc.f90
+++ b/src/acefc.f90
@@ -7156,7 +7156,7 @@ contains
    integer::last,nx,ix
    integer::jp,jpn,jpp
    real(kr)::test,eemx,yield,xnext,xx,yy,y,xn,eyl,gyl,en
-   real(kr)::apsx,step1,step2,xl,pl,yn,pn,rn,sum,ee
+   real(kr)::apsx,step1,step2,xl,pl,yn,pn,rn,sum,ee,eout1
    real(kr)::ep,e,bzro,sfe,sfo,bbi,fbarcm,delfcm,akal,rkal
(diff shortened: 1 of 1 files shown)


whaeck:
@HunterBelanger I'm changing the branch to merge into to a temporary branch so I can update the release notes at the same time. Once I update the release notes, I'll merge this into develop

PR #406, opened 2026-07-14
Read ENDF/B-TAB1 format and dynamically allocate

wcdawn:
In a particular applicatin of NJOY, the `ntmp` value in `getwtf` had been increased to allow reading longer arrays of weight functions. In the ENDF/B-TAB1 format, the length of the array is specified as part of the input file.

This PR introduces a subroutine `rdtabl` in `util.f90` which first reads the headder information of the table (first six entries), allocates the necessary space, uses `backspace`, and then reads the array as appropriate.

This new subroutine could be useful elsewhere, but I've only used it in `getwtf` in `groupr.f90` for now.


src/util.f90
```diff
@@ -464,5 +465,36 @@ subroutine a10(x,hx)
...
+   np = nint(dmy(6))
+   arrlen = 6 + 2*nr + 2*np
+   
+   if (allocated(arr)) then
+      deallocate(arr)
+   end if
+   allocate(arr(arrlen))
+
+   backspace(iunit)
+   read(iunit, *) (dmy(i),i=1,arrlen)
```
HunterBelanger:
```suggestion
   read(iunit, *) (arr(i),i=1,arrlen)
```
I think this should be `arr` instead of `dmy` ?

wcdawn: Good catch. Thanks!

src/groupr.f90
```diff
@@ -5119,7 +5113,7 @@ subroutine getwtf(e,enext,idis,lord,wtf)
    !-------------------------------------------------------------------
    use physics ! provides bk
    use endf    ! provides terpa
-   use util    ! provides sigfig
+   use util    ! provides sigfig,rdtabl
```
HunterBelanger: Also, I think this comment modification should be on the previous subroutine instead of this one ?

wcdawn: Yep. Fixed.

PR #408, opened 2026-07-23
Ensure Release builds are optimised

shimwell:
Just another small PR for your consideration

Some toolchains and CMake presets leave CMAKE_Fortran_FLAGS_RELEASE empty, which produces an unoptimised (-O0) executable even for a Release build - several times slower with identical results.

Inject -O3 for Release only when no optimisation level has already been requested, so user-provided flags still take precedence.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,15 @@ set( CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "Supported configura
 
 project( njoy LANGUAGES Fortran )
 
+# Guarantee that a Release build is actually optimised. Some toolchains and CMake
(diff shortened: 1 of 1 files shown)


wcdawn:
This is a nice catch! I didn't personally notice a measurable time difference between the runtime of the default compilation and the optmized compilation. Do you have an example where it is significant?

Also, I noticed that this only is triggered if `-DCMAKE_BUILD_TYPE=Release` is passed. By default, this is not the case. Something to consider. I think that the optimized build should probably be default behavior.

shimwell:
When running the [openmc data generation scripts](https://github.com/openmc-dev/data) for producing complete libraries (endf to hdf5) I have noticed it is quicker

PR #409, opened 2026-07-23
Fix/ismooth non zero

whaeck:
Adding the fix contributed by @HunterBelanger and updating the release notes.


HunterBelanger:
Also, this closes #277 and perhaps #131 as well ?

whaeck:
Yes, I think we can do that.

